PR #20, opened 2019-11-21
I003 inventory

thuemmels:
Adds an inventory as a game object with the Player object as the parent.
Adds items in form of scriptable objects with a name and a sprite.


Assets/Scripts/Item.cs
```diff
@@ -0,0 +1,12 @@
...
+using System.Collections.Generic;
+using UnityEngine;
+
+[CreateAssetMenu(fileName = "New Item", menuName = "Inventory/Item")]
+public class Item : ScriptableObject
+{
+    [SerializeField]
+    new private string name = "New Item";
+    [SerializeField]
+    private Sprite icon = null;
```
DavidPgl, line 9: I guess those fields should be public. Otherwise the inventory can't access the correct sprite and name.
FYI: Default value "null" is always the case, so no need to define it with it. If we decide to always define fields, then it's fine, but those values will always be overwritten by the values from the inspector which can be a bit confusing.

thuemmels: You're right. Or Getter/Setter for each field. Which one do you prefer?

DavidPgl: Probably a getter only with a property:
`public Sprite Icon => icon;`
Also it's nicer to give the file the specific item name and don't use an extra one. Easier to find later on

thuemmels: > Also it's nicer to give the file the specific item name and don't use an extra one. Easier to find later on

The idea was that this item script is for general purpose.
Each specific item (such as wood) is stored under the Items folder. Is that what you meant?

DavidPgl: I meant that each ScriptableObject should be named after the item itself. So getting the name of the ScriptableObject itself will give you the item name. So there is no need for a field.

Assets/Scripts/Inventory.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class Inventory : MonoBehaviour
+{
+    [SerializeField]
+    private List<Item> items = new List<Item>();
+
+    public void Add(Item item)
```
Kaes3kuch3n: Just as a hint :) Methods like these can be simplified to:
```c#
public void Add(Item item) => items.Add(item);
```

Assets/Scripts/Item.cs
Kaes3kuch3n: The existing field `name`, which you are overriding, contains the name of the actual file which represents the instance of the scriptable object 👍

thuemmels: Ah, got it, yeah that's probably better :)

Assets/Scripts/Inventory.cs
thuemmels: Thanks, I've included that.

Kaes3kuch3n (approved):

SamuelllG (approved):

PR #23, opened 2019-11-23
Attack system

SamuelllG:
- implemented reusable attack system for player and enemies
- adapted playercontroller and enemycontroller to use attacklogic
- implemented IDamageable and IMovable interface (IMovable could also be an abstract class and inherit 4-5 methods which are the same for player and enemy)
- implemented hit marker / animation replacement as material (color) change (this is in Stats class which will be replaced with real stats class)


Assets/Scripts/AttackLogic.cs
```diff
@@ -0,0 +1,175 @@
...
+    // ----temporary as animation replacement------
+        private MeshRenderer _gameObjectRenderer;
+        private Material _prevMat;
+        private Material _attackAnimationMaterial;
+    // ---------
+
+    private void Awake()
+    {
+        // init components
+        _movable = GetComponent<IMovable>();
```
DavidPgl, line 51: Always a good idea to add `[RequiresComponent(typeOf(IMovable))` to the class, if you use GetComponent in the Awake

Assets/Scripts/AttackLogic.cs
```diff
@@ -0,0 +1,175 @@
...
+
+        // -----temp replacement for animation-----
+            _gameObjectRenderer = gameObject.GetComponentInChildren<MeshRenderer>();
```
DavidPgl, line 54: Please use a reference:(

Assets/Scripts/AttackLogic.cs
```diff
@@ -0,0 +1,175 @@
...
+        else if (PerformingHit)
+        {
+            bool? success = PerformHit();
+            if (success != null)
+            {
+                // hit animation performed
+                if (success == true)
+                {
+                    // deal damage
+                    IDamageable damageable = (IDamageable)_target.GetComponent<IDamageable>();
```
DavidPgl: You could define _target as IDamageable and save one GetComponent call. GetComponent is very heavy. If you can avoid it, try to.

Assets/Scripts/AttackLogic.cs
```diff
@@ -0,0 +1,175 @@
...
+    public void StopAttack()
+    {
+        _target = null;
+    }
+
+    /// <summary>
+    /// Waits (attackTime) for hit to be finished and returns the result.
+    /// </summary>
+    /// <returns>Null if hit not finished, true if hit successful performed, false if unsuccessful</returns>
+    private bool? PerformHit()
```
DavidPgl: Probably personal preference but not the biggest fan of making bool to a three-state type. This is probably one of the few cases where using it is still readable. Defining three Enums for this would probably look nicer though:
`AttackStatus.Hit`
`AttackStatus.Miss`
`AttackStatus.NotFinished`
or something like that.
Also preferences but Unity offers **Coroutines**, in which you can wait for a time or a specified event to be completed. But I'm not sure if you can use it here

Assets/Scripts/AttackLogic.cs
```diff
@@ -0,0 +1,175 @@
...
+    {
+        _timer += Time.deltaTime;
+
+        if (_timer > attackTime)
+        {
+            // animation finished
+            _timer = 0;
+            // target still in range?
+            float distanceToTarget = Vector3.Distance(_target.transform.position, transform.position);
```
DavidPgl: You calculate the distance each frame anywhere (line 78), so saving it in a variable would save you one calculation

Assets/Scripts/Interfaces/IDamageable.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Interfaces
+{
+    public interface IDamageable
```
DavidPgl: Would an abstract class be useful here? I mean enemies and the player both share health and the "loosing health behavior".

Assets/Scripts/Interfaces/IMovable.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Interfaces
+{
+    public interface IMovable
```
DavidPgl, line 7: Same as above. Only if the methods do the same thing of course.

Assets/Scripts/Player/PlayerController.cs
```diff
@@ -83,30 +102,106 @@ private void Update()
...
+            if (Physics.Raycast(clickRay, out RaycastHit hit, 10000) && !_attackLogic.PerformingHit)
             {
-                // Set destination for nav mesh agent
-                _navMeshAgent.SetDestination(hit.point);
-                _navMeshAgent.isStopped = false;
-
-                // Create click point effect
-                Instantiate(clickEffect, hit.point + Vector3.up * 5, Quaternion.identity);
+                GameObject objectHit = hit.collider.gameObject;
+                IDamageable damageable = (IDamageable)objectHit.GetComponent<IDamageable>();
```
DavidPgl: Is the cast really necessary?

Assets/Scripts/Player/PlayerController.cs
```diff
@@ -83,30 +102,106 @@ private void Update()
...
+        /// <param name="target">The target to face</param>
+        /// <param name="shouldTurn">Whether the object should turn to the target or not</param>
+        /// <returns>Whether the object is facing the target</returns>
+        public bool FaceTarget(GameObject target, bool shouldTurn)
+        {
+            float f;
+            return FaceTarget(target, shouldTurn, out f);
+        }
+
+        // ----- Note: same in EnemyController --> make IMovable abstract class and inherit? ------
```
DavidPgl, line 180: Sounds good to me

Assets/Scripts/Player/PlayerController.cs
```diff
@@ -83,30 +102,106 @@ private void Update()
...
+        /// <summary>
+        /// Faces the target. Returns true if facing the target.
+        /// </summary>
+        /// <param name="target">The target to face</param>
+        /// <param name="shouldTurn">Whether the object should turn to the target or not</param>
+        /// <returns>Whether the object is facing the target</returns>
+        public bool FaceTarget(GameObject target, bool shouldTurn)
+        {
+            float f;
+            return FaceTarget(target, shouldTurn, out f);
```
DavidPgl:
```suggestion
            return FaceTarget(target, shouldTurn, out _);
```
Then you can delete the local variable

Assets/Scripts/Player/Stats.cs
```diff
@@ -0,0 +1,64 @@
+﻿using Interfaces;
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
```
DavidPgl: Namespace missing

Assets/Scripts/Player/Stats.cs
```diff
@@ -0,0 +1,64 @@
...
+        // just put initial mat here
+        _prevMat = _gameObjectRenderer.material;
+    }
+
+    /// <summary>
+    /// Changes the objects color back to normal after being hit.
+    /// </summary>
+    private void Update()
+    {
+        if (_hit)
```
DavidPgl:
```suggestion
        if (!_hit) return
```
This way it will skip the rest

Assets/Scripts/AttackLogic.cs
```diff
@@ -0,0 +1,175 @@
...
+            // chase target
+            _movable.Move(_target.transform.position);
+        }
+    }
+
+    /// <summary>
+    /// Starts the attack.
+    /// </summary>
+    /// <param name="target">The target to be attacked</param>
+    public void StartAttack(GameObject target)
```
DavidPgl, line 157:
```suggestion
    public void StartAttack(IDamageable target)
```
Make it as specific as possible

DavidPgl (changes requested):

thuemmels (dismissed):

Assets/Scripts/AttackLogic.cs
SamuelllG: over inspector? but it's just temporary...

Assets/Scripts/AttackLogic.cs
SamuelllG: How do I save one GetComponent call then? the only point where I call GetComponent is when the hit was performed successfully, if I define _target as IDamageable, I'd have to GetComponent everytime independent of the hit being successful or not. Also this would cause trouble at different places like calculating the distance etc.

Assets/Scripts/AttackLogic.cs
SamuelllG: as stated above, same thing

Assets/Scripts/AttackLogic.cs
SamuelllG: Done, but doesn't improve readability in my opinion. Also removed PerformingHit variable and put it all in one: Status.

Assets/Scripts/AttackLogic.cs
SamuelllG: Indeed, however I always try to keep the number of instance variables low (matter of readability), and a parameter doesn't really make sense. But I guess, if it's better to make one, I'll make one ^^

Assets/Scripts/Interfaces/IDamageable.cs
SamuelllG: I thought about that as well, maybe we have to discuss this. In my opinion not a matter of this PR, we can create a separate issue for this :)

Assets/Scripts/Interfaces/IMovable.cs
SamuelllG: yep

Assets/Scripts/Player/PlayerController.cs
SamuelllG: nope xD

Assets/Scripts/Player/PlayerController.cs
SamuelllG: we should a separate issue for this

Assets/Scripts/Player/Stats.cs
SamuelllG: yep, I will integrate this into PlayerState

Assets/Scripts/Player/Stats.cs
SamuelllG: and the advantage is? skip one if statement? I rather take the readability, thanks 😄 (I changed it a bit)

SamuelllG:
I also merged master which was a bit of a pain 😅 and fixed everything. I added the abstract class discussion as a task as well as implementing trigger colliders for the NPCs :)

Assets/Scripts/AttackLogic.cs
Kaes3kuch3n: For temporary stuff I'd suggest adding a //TODO or //BUG so that we can find stuff like this easily later

Assets/Scripts/AttackLogic.cs
```diff
@@ -0,0 +1,196 @@
...
+    /// <summary>
+    /// Ends the hit. Deals damage to the target, if hit was successful. Resets aggro if necessary.
+    /// </summary>
+    private void HitFinished()
+    {
+        // hit animation performed
+        if (Status == AttackStatus.Hit)
+        {
+            // deal damage
+            IDamageable damageable = (IDamageable)_target.GetComponent<IDamageable>();
```
Kaes3kuch3n: I'm pretty sure this cast ist redundant, since GetComponent<IDamageable> returns an object of type IDamageable 🤔

Assets/Scripts/AttackLogic.cs
Kaes3kuch3n: Could you please elaborate on that? It's not clear at all to me what you mean by that/what you're referring to 😅 

Reply:
You sound like Prof. WeWu :D
Since I couldn't reply, here with an edit: I added a few words to the summary. "If the target is not in attacking range, chases it. If the target is near performs a hit on it and either resets or continues attacking." -- just the normal "attack cycle".

Assets/Scripts/EnemyController.cs
```diff
@@ -150,27 +164,65 @@ private void MoveToNextPoint()
...
+    /// <summary>
+    /// Faces the target. Returns true if facing the target.
+    /// </summary>
+    /// <param name="target">The target to face</param>
+    /// <param name="shouldTurn">Whether the object should turn to the target or not</param>
+    /// <returns>Whether the object is facing the target</returns>
+    public bool FaceTarget(GameObject target, bool shouldTurn)
+    {
+        float f;
+        return FaceTarget(target, shouldTurn, out f);
```
Kaes3kuch3n: If you don't wanna use an out variable, you can simply use `out _` to discard the result

Assets/Scripts/Interfaces/IMovable.cs
```diff
@@ -0,0 +1,17 @@
...
+    {
+        void Move(Vector3 position);
+
+        bool FaceTarget(GameObject target, bool shouldTurn, out float difference);
+
+        bool FaceTarget(GameObject target, bool shouldTurn);
```
Kaes3kuch3n: Feels kinda redundant with the other FaceTarget method defined right above it, since we can just discard the out value if we don't need it

Assets/Scripts/Player/PlayerController.cs
```diff
@@ -114,26 +132,101 @@ private void SetUpControls()
...
         /// </summary>
-        private void Move(InputAction.CallbackContext obj)
+        private void OnRightClick(InputAction.CallbackContext obj)
         {
             Ray clickRay = _camera.ScreenPointToRay(Mouse.current.position.ReadValue());
 
-            // Get ground position from mouse click
-            if (Physics.Raycast(clickRay, out RaycastHit hit, 10000, moveClickLayers))
+            // only change target / move, if not performing a hit
+            if (Physics.Raycast(clickRay, out RaycastHit hit, 10000) && _attackLogic.Status == AttackLogic.AttackStatus.None)
```
Kaes3kuch3n, line 145: Would maybe change this in the future to make it possible to cancel attacks by moving away from the enemy. Doesn't really feel like the player is in control like this 😬

Assets/Scripts/Player/PlayerController.cs
```diff
@@ -154,11 +247,11 @@ private void RotateCamera(InputAction.CallbackContext obj)
...
         private void UpdateCameraAngle(float cameraDistanceChange = 0)
         {
-            float radian = (float) Math.PI * _cameraAngleX / 180;
+            float radian = (float)Math.PI * _cameraAngleX / 180;
             cameraDistance.y = Mathf.Clamp(cameraDistance.y + cameraDistanceChange, cameraDistanceRange.min,
                 cameraDistanceRange.max);
-            _cameraPosition = new Vector3((float) Math.Sin(radian) * cameraDistance.x, cameraDistance.y,
-                (float) -Math.Cos(radian) * cameraDistance.x);
+            _cameraPosition = new Vector3((float)Math.Sin(radian) * cameraDistance.x, cameraDistance.y,
+                (float)-Math.Cos(radian) * cameraDistance.x);
```
Kaes3kuch3n, line 243: Why did you remove the spaces? 😂

Assets/Scripts/PlayerManager.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class PlayerManager : MonoBehaviour
+{
+    #region Singleton
```
Kaes3kuch3n: Could you switch to David's Singleton base class for this logic?

Kaes3kuch3n (changes requested):

Assets/Scripts/AttackLogic.cs
DavidPgl: It only takes a minute. Don't be lazy:P

Assets/Scripts/AttackLogic.cs
SamuelllG: for sure, ye I forgot that here

Assets/Scripts/EnemyController.cs
SamuelllG: forgot that here as well

Assets/Scripts/Interfaces/IMovable.cs
SamuelllG: good point

Assets/Scripts/Player/PlayerController.cs
SamuelllG: We can discuss about that. It might be useful, idk. It's a tiny change howsoever.

Assets/Scripts/Player/PlayerController.cs
SamuelllG: I didn't, must have come from merging.

Assets/Scripts/PlayerManager.cs
SamuelllG: didn't see there is already one - switched

Kaes3kuch3n (dismissed):

Assets/Scripts/AttackLogic.cs
DavidPgl: _target is of type GameObject and will be set at some point. Is there a reason why you can't define it as a IDamageable from the beginning? Attacking will happen more often than setting the _target, right?

Assets/Scripts/EnemyController.cs
```diff
@@ -47,9 +52,13 @@ public class EnemyController : MonoBehaviour
     [SerializeField]
     private bool showEscapeRadius;
 
-    private Transform _target;
+    // component references
     private NavMeshAgent _agent;
     private WanderAI _wanderAI;
+    private AttackLogic _attackLogic;
```
DavidPgl, line 58: So this script accesses AttackLogic and AttackLogic accesses the EnemyController (with IMovable). Does the AttackLogic really care about IMovable? Wouldn't it be enough if the EnemyController activates the attacking and stops moving?

Assets/Scripts/Managers/PlayerManager.cs
```diff
@@ -1,21 +1,16 @@
...
-using UnityEngine;
-using UnityEngine.Networking.PlayerConnection;
+﻿using UnityEngine;
 
 namespace Managers
 {
     public class PlayerManager : Singleton<PlayerManager>
     {
-        [SerializeField] private GameObject player;
+        [SerializeField] private GameObject _player;
```
DavidPgl: SerializedFields are normaly without underscore. Was already correct:)

Assets/Scripts/Player/EnemyState.cs
```diff
@@ -0,0 +1,63 @@
...
+/// </summary>
+public class EnemyState : MonoBehaviour, IDamageable
+{
+    [Tooltip("The time the object should be marked as hit after being hit")]
+    [SerializeField]
+    private float hitMarkTime;
+
+    [Tooltip("The MeshRenderer of the graphics object of the player")]
+    [SerializeField]
+    private MeshRenderer _gameObjectRenderer;
```
DavidPgl: Stick with SerializedField without underscore

Assets/Scripts/Player/PlayerState.cs
```diff
@@ -21,7 +21,24 @@ public class PlayerState : MonoBehaviour, IDamageable
...
+
+
+        // ------temp for hit animation------
+        [Tooltip("The time the object should be marked as hit after being hit")]
+        [SerializeField]
+        private float hitMarkTime;
+
+        [Tooltip("The MeshRenderer of the graphics object of the player")]
+        [SerializeField]
+        private MeshRenderer _gameObjectRenderer;
```
DavidPgl:
```suggestion
        private MeshRenderer gameObjectRenderer;
```

Assets/Scripts/Player/EnemyState.cs
DavidPgl:
```suggestion
    private MeshRenderer gameObjectRenderer;
```

Assets/Scripts/Managers/PlayerManager.cs
DavidPgl:
```suggestion
        [SerializeField] private GameObject player;
```

DavidPgl:
Das ist das letzte :P

Assets/Scripts/Managers/PlayerManager.cs
SamuelllG: oh ye sure :joy: my bad

Assets/Scripts/Player/EnemyState.cs
SamuelllG: dang, what did I do

Assets/Scripts/Managers/PlayerManager.cs
SamuelllG: for a singleton I don't expect a serialize

Assets/Scripts/EnemyController.cs
SamuelllG: That's a good point, but AttackLogic needs IMovable anyway, so that it has a target which it can turn too and which it can check the position after completing the hit :)

I look at it like this: 
You click on the ground --> you move --> Controller
You click on an enemy --> you attack --> AttackLogic
If you have to move in order to attack the enemy that's part of the AttackLogic. Otherwise you'd have to implement something like "target out of range" when clicking on an enemy far away.

DavidPgl: Ahh I see, I was always thinking this way:
You click on something: ground or enemy or resource (Not in range) -->Move to it --> Movable
You click on something (In range) --> Trigger action (e.g. attack or collect) --> specific script/component
Both works I guess

DavidPgl (approved):

Assets/Scripts/AttackLogic.cs
SamuelllG: haha you're talking about two different things ^^
I made it serializable, Luis talks about the whole animation / temporary color thingy

Assets/Scripts/AttackLogic.cs
SamuelllG: I don't know if setting the target (clicking on the enemy) will happen less often than a successful hit... especially because I know people that rape the mouse button on attack :D

(I need the gameobject so that I have the position of the enemy and that I can calculate the distance.)

Assets/Scripts/AttackLogic.cs
SamuelllG: Oh okay it also didn't show me that you commented on existing comments... what I'm talking about is the discussion of whether to use IDamageable as target or GameObject (see the only open discussion above)

Kaes3kuch3n (approved):

PR #24, opened 2019-11-24
Player state

inkompotato:
Implemented a PlayerState that keeps the values for health, hunger and thirst. This class also invokes Events each time a state updates. Other components are able to listen for these events, e.g. UI. 

Implemented a simple remote logging class that has no current purpose except testing. Can connect to a TCP Server.

known issues:
- scene file changed
- ProjectVersion.txt, sorry wanted to try something in the beta...


Assets/Scripts/Player/PlayerController.cs
```diff
@@ -104,6 +110,8 @@ private void Move(InputAction.CallbackContext obj)
 
                 // Create click point effect
                 Instantiate(clickEffect, hit.point + Vector3.up * 5, Quaternion.identity);
+                
+                playerPositionUpdated.Invoke(transform.position);
```
DavidPgl: If I understand this correctly, then it will invoke the event after the player clicks on the screen but before the player reaches the new destination, right? So this will give back the position before the player has moved. Is that wanted?

Assets/Scripts/Player/PlayerState.cs
```diff
@@ -0,0 +1,78 @@
+﻿using System;
+using UnityEngine;
+using UnityEngine.Events;
+using Random = System.Random;
+
+namespace Player
+{
+    
+    [System.Serializable] public class UnityEventInt:UnityEvent<int> {}
```
DavidPgl: Never saw this before and don't know about performance, but why not use the standard Event-System with:
`public delegate void PlayerStateChanged(int newValue)`
`public event PlayerStateChanged OnPlayerHealthChanged`
`public event PlayerStateChanged OnPlayerHungerChanged`

Assets/Scripts/Player/PlayerController.cs
DavidPgl: Is this even used?

Assets/Scripts/Remote/RemoteStatusHandler.cs
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        private void Disconnect()
+        {
+            _stream.Close();
+            _client.Close();
+        }
+        
+        //event listeners
+        public void HeathUpdated(int newValue)
```
DavidPgl: Typo

Assets/Scripts/Player/PlayerController.cs
inkompotato: it only sends the position at every click for better performance, currently not really used, only for remote logging which is experimental

Assets/Scripts/Player/PlayerState.cs
inkompotato: the tutorial I used suggested this ¯\_(ツ)_/¯

Assets/Scripts/Remote/RemoteStatusHandler.cs
inkompotato: just looked at this for 3min before I realized whats wrong

Assets/Scripts/Player/PlayerState.cs
DavidPgl: Wieder was gelernt :o

Assets/Scripts/Player/PlayerController.cs
DavidPgl: Not really used, which means it can be removed? 👀

inkompotato: Nope, I use remote Services to monitor in game events like position change and state change...

Assets/Scripts/Remote/RemoteStatusHandler.cs
```diff
@@ -0,0 +1,124 @@
...
+        
+        private TcpClient _client;
+        private NetworkStream _stream;
+
+        [SerializeField] private string ip;
+        [SerializeField] private int port;
+        [SerializeField] private bool use;
+        
+        [Header("Event that will throw when the player's health has been changed via Remote")]
+        public UnityEventInt playerHeathRemoteUpdate;
```
DavidPgl: Typo:P

Assets/Scripts/UI/StatBar.cs
```diff
@@ -0,0 +1,93 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using UnityEditor;
+using UnityEngine;
+using UnityEngine.UI;
+
+public class StatBar : MonoBehaviour
+{
+    public string SpriteLocation = "Sprites/PlayerHungerStates_Sprite";
```
DavidPgl: Please use:
`[SerializedField] private Sprite[] playerHungerStateSprites;`

Assets/Scripts/UI/StatBar.cs
```diff
@@ -0,0 +1,93 @@
...
+
+    private Sprite[] _barStates;
+
+    private Image[] _images;
+
+    private Animator _animator;
+    
+    
+    // Start is called before the first frame update
+    void Start()
```
DavidPgl: Please define an access modifier. I know it's not needed but it looks nicer and you know, Consistency 🎉

Assets/Scripts/UI/StatBar.cs
```diff
@@ -0,0 +1,93 @@
...
+    {
+        _barStates = Resources.LoadAll<Sprite>(SpriteLocation);
+    }
+
+    // Update is called once per frame
+    void Update()
```
DavidPgl: Remove if not needed

Assets/Scripts/UI/StatBar.cs
```diff
@@ -0,0 +1,93 @@
...
+    {
+         
+    }
+
+    public void UpdateBar(int value)
+    {
+        var newValue = value;
```
DavidPgl: Don't hate "var" anymore but again, it should be consistent. Please use specific types

Assets/Scripts/UI/StatBar.cs
```diff
@@ -0,0 +1,93 @@
...
+        var oldValue = _statusValue;
+
+        if (newValue > 100 || newValue < 0)
+        {
+            Debug.Log("Status bar value has to be between 0 and 100");
+        }
+        else if (newValue == oldValue)
+        {
+            Debug.Log("new status bar value has to differ from the old value");
```
DavidPgl: That shouldn't be an error. This would allow us to animate the bar, e.g. highlight it, even though the player had full hunger/thirst

Assets/Scripts/UI/StatBar.cs
DavidPgl: Also why not use value for everything? And _statusValue

Assets/Scripts/UI/StatBar.cs
```diff
@@ -0,0 +1,93 @@
...
+            _statusValue = newValue;
+            ChangeSprite();
+        }
+    } 
+    
+    private void ChangeBarAnimation(int newValue, int oldValue)
+    {
+        if (newValue > oldValue)
+        {
+            _animator = gameObject.GetComponent<Animator>();
```
DavidPgl: Make a reference and get it on Awake:)

Assets/Scripts/UI/StatBar.cs
```diff
@@ -0,0 +1,93 @@
...
+            {
+                _animator.Play("UIDecreaseStat");
+
+            }
+        }
+    }
+
+    private void ChangeSprite()
+    {
+        if (_statusValue >= 0 && _statusValue < 10) gameObject.GetComponent<Image>().sprite = _barStates[0];
```
DavidPgl: Same here, make a reference to the image

Assets/Scripts/UI/StatBar.cs
DavidPgl: Or SerializedField and link it in the inspector

Assets/Scripts/Player/PlayerController.cs
```diff
@@ -1,11 +1,14 @@
 ﻿using System;
 using UnityEngine;
 using UnityEngine.AI;
+using UnityEngine.Events;
 using UnityEngine.InputSystem;
 using Utils;
 
 namespace Player
 {
+    [System.Serializable] public class UnityEventVector3:UnityEvent<Vector3> {}
```
Kaes3kuch3n: Why is this a separate class? Is the base class (UnityEvent<T>) not serialisable?

Assets/Scripts/Player/PlayerState.cs
Kaes3kuch3n: Same question as above :)

Assets/Scripts/Player/PlayerState.cs
```diff
@@ -0,0 +1,78 @@
...
+    public class PlayerState : MonoBehaviour
+    {
+        private readonly Random _random = new Random();
+        
+        //Player State values
+        [Tooltip("100 - full health, 0 - dead")] [SerializeField] [Range(0, 100)] 
+        private int health;
+
+        [Tooltip("100 means no hunger, 0 means the player is in desperate need for food")] [SerializeField] [Range(0, 100)] 
+        private int hunger;
```
Kaes3kuch3n: I would rename it to saturation, if the scale ist 100 = full, 0 = empty

Assets/Scripts/UI/StatBar.cs
DavidPgl: FYI: You can use this for multiple references:
https://gyazo.com/4fea78d861640bbb93918df96a4dc0d6

Assets/Scripts/Player/PlayerState.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        [Tooltip("100: not thirsty, 0: gazing for a sip of water")] [SerializeField] [Range(0, 100)] 
+        private int thirst;
```
Kaes3kuch3n: I would rename it to hydration, if the scale ist 100 = full, 0 = empty

Assets/Scripts/UI/StatBar.cs
```diff
@@ -0,0 +1,93 @@
...
+        if (_statusValue >= 10 && _statusValue < 20) gameObject.GetComponent<Image>().sprite = _barStates[1];
+        if (_statusValue >= 20 && _statusValue < 30) gameObject.GetComponent<Image>().sprite = _barStates[2];
+        if (_statusValue >= 30 && _statusValue < 40) gameObject.GetComponent<Image>().sprite = _barStates[3];
+        if (_statusValue >= 40 && _statusValue < 50) gameObject.GetComponent<Image>().sprite = _barStates[4];
+        if (_statusValue >= 50 && _statusValue < 60) gameObject.GetComponent<Image>().sprite = _barStates[5];
+        if (_statusValue >= 60 && _statusValue < 70) gameObject.GetComponent<Image>().sprite = _barStates[6];
+        if (_statusValue >= 70 && _statusValue < 80) gameObject.GetComponent<Image>().sprite = _barStates[7];
+        if (_statusValue >= 80 && _statusValue < 90) gameObject.GetComponent<Image>().sprite = _barStates[8];
+        if (_statusValue >= 90 && _statusValue < 100) gameObject.GetComponent<Image>().sprite = _barStates[9];
+        if (_statusValue == 100) gameObject.GetComponent<Image>().sprite = _barStates[10];
```
DavidPgl: This is not really adaptive. If the length of the array changes. this could break. Also this will go through each if-statement every time (if the player's statusValue is high/full). I would either try to come up with a nice way for keeping track of the "levels" and the corresponding sprite or simply calculate the index. If the steps stay the same you could do something similar to this:
`int index = thirst / sprites.Length;`
`Sprite newSprite = sprites[index];`
Probably with a better calculation 😅 
If you want some more control over the step-size of the levels you need write a fancier calculation though

DavidPgl (changes requested):

Assets/Scripts/Player/PlayerState.cs
```diff
@@ -0,0 +1,78 @@
...
+        
+        [Header("Event that will throw when the player's health changes")]
+        public UnityEventInt playerHeathUpdated;
```
Kaes3kuch3n: Typo :P

Assets/Scripts/Player/PlayerState.cs
```diff
@@ -0,0 +1,78 @@
...
+            else if (updatedValue < 0)
+            {
+                updatedValue = 0;
+                ChangePlayerHealth(-1);
+            }
+
+            thirst = updatedValue;
+            playerThirstUpdated.Invoke(updatedValue);
+        }
+    }
```
Kaes3kuch3n, line 90: I think it would make sense to separate these into two methods each, e.g. `void Heal(int amount)` & `void Damage(int amount)`. Makes the intention clearer when calling the methods and breaks them down into smaller, easier maintainable parts

Assets/Scripts/UI/StatBar.cs
```diff
@@ -0,0 +1,93 @@
...
+using UnityEngine;
+using UnityEngine.UI;
+
+public class StatBar : MonoBehaviour
+{
+    public string SpriteLocation = "Sprites/PlayerHungerStates_Sprite";
+    
+    private int _statusValue = 100;
+
+    private Sprite[] _barStates;
```
Kaes3kuch3n: This can be a public field → Sprites can be dragged into the array in the inspector, so the sprite location is not needed

Assets/Scripts/UI/StatBar.cs
Kaes3kuch3n: Please remove code that's not needed :)

Assets/Scripts/UI/StatBar.cs
Kaes3kuch3n: Please do not use var → see code conventions (and I think this variable is not really needed at all 😅)

Assets/Scripts/UI/StatBar.cs
Kaes3kuch3n: Please use else ifs or add a return statement to each if, otherwise all of them are checked every time

Kaes3kuch3n (changes requested):

Assets/Scripts/Player/PlayerController.cs
inkompotato: suggested by tutorial I did

Assets/Scripts/Player/PlayerState.cs
inkompotato: same answer as above :)

Assets/Scripts/UI/StatBar.cs
```diff
@@ -19,21 +19,15 @@ public class StatBar : MonoBehaviour
...
-    {
-         
-    }
-
+    
     public void UpdateBar(int value)
     {
-        var newValue = value;
-        var oldValue = _statusValue;
+        int newValue = value;
```
Kaes3kuch3n: Still not really needed, I think :P I'd suggest renaming `value` → `newValue`

Assets/Scripts/Player/PlayerState.cs
inkompotato: I think a positive value for heal and a negative value for damage is clear enough. The methods are small enough and do not need to be broken up for maintainability reasons.

Assets/Scripts/UI/StatBar.cs
inkompotato: serialized, unity editor is annoying, took a while until I found out that you can lock panels

Assets/Scripts/UI/StatBar.cs
inkompotato: removed

Assets/Scripts/UI/StatBar.cs
inkompotato: now on awake

Assets/Scripts/UI/StatBar.cs
inkompotato: now reference on awake

Assets/Scripts/UI/StatBar.cs
inkompotato: `_image.sprite = barStates[_statusValue/10];`, we do not expect the array to change

Assets/Scripts/Player/PlayerController.cs
inkompotato: leaving it as is

Assets/Scripts/Player/PlayerState.cs
inkompotato: leaving it as is

Assets/Scripts/Player/PlayerState.cs
inkompotato: no changes for now

Assets/Scripts/UI/StatBar.cs
inkompotato: serialized, see further up

Assets/Scripts/UI/StatBar.cs
inkompotato: replaced with formula, see further up

Assets/Scripts/UI/StatBar.cs
inkompotato: renamed

Assets/Scripts/Player/PlayerController.cs
Kaes3kuch3n: Okay, found out that the generic base class ist not serialisable 😬 Then I'd suggest we either don't use a serialised version, but set the event subscriptions through code instead, or we create these events in separate files, e.g. in the Utils.Events namespace, and call them something like Vector3Event, IntEvent or so

Assets/Scripts/Player/PlayerState.cs
SamuelllG: I agree with Jan. I think an interface IDamageable would be handy. If we make PlayerState implement IDamageable, we should add another method like Hit(int damage) which itself calls ChangePlayerHealth with (-damage). Also you could give the interface a method Heal(int amount) which does the same just with healing, because damageables should also be able to be healed, so this fits in here.

Also see implementation of attack system.

SamuelllG: + ChangePlayerHealth = private ;)

inkompotato:
All requested changes are now addressed

DavidPgl (dismissed):

Kaes3kuch3n (dismissed):

DavidPgl (approved):

Kaes3kuch3n (approved):

PR #29, opened 2019-11-30
Interaction

cuongtm95:
### Script for all interactable objects in the game (like items, enemies, animals, treasure, etc..)

- [x] **Created a Interactable Base script called "Interactable", that will do the following:**

- Character will stop at a distance in front of the object (while facing it) to perform action (instead of going through/inside it)
- There's a **virtual** function inside called **Interact()** which is meant to be overwritten from other scripts to perform a specific action (For example: Open crate, Chop tree, Cook food, Make fire...)

- [x] **Pick up item and add to Inventory**

- [ ] **Show tooltip when hover over objects**

- [ ]  ...


SamuelllG:
Why are @Kaes3kuch3n 's commits in here? We have a separate PR for this

SamuelllG (changes requested):
Please have a look at our coding conventions here https://paper.dropbox.com/doc/General-Project-Conventions--ApoocCghuHUWZggkaloUj0IZAg-Z8kbXCDr3XqW11DkCtUHe or look at approved PRs and change a few things in your code. 

Please make Interactable an abstract class / interface. 
For the PlayerController, there is new stuff similar to yours (**as soon as someone approves the Attack_System PR**) that you can integrate the Interactable stuff into (see OnRightClick method). You'll have to pull master into your branch and battle with the merge conflicts :)
Methods for chasing enemies, IMovable etc. are already done. I guess items won't move. Depending on how much "logic" we need for interacting, you could create a class like InteractLogic, that you add to the player and call in PlayerController's OnRightClick (mentioned above). For now it's really just picking the item up, isn't it? So for now it might be enough to do that in a short method in PlayerController.

cuongtm95:
> Please have a look at our coding conventions here https://paper.dropbox.com/doc/General-Project-Conventions--ApoocCghuHUWZggkaloUj0IZAg-Z8kbXCDr3XqW11DkCtUHe or look at approved PRs and change a few things in your code.
> 
> Please make Interactable an abstract class / interface.
> For the PlayerController, there is new stuff similar to yours (**as soon as someone approves the Attack_System PR**) that you can integrate the Interactable stuff into (see OnRightClick method). You'll have to pull master into your branch and battle with the merge conflicts :)
> Methods for chasing enemies, IMovable etc. are already done. I guess items won't move. Depending on how much "logic" we need for interacting, you could create a class like InteractLogic, that you add to the player and call in PlayerController's OnRightClick (mentioned above). For now it's really just picking the item up, isn't it? So for now it might be enough to do that in a short method in PlayerController.

Gotcha, gonna deal with it right away.

SamuelllG:
Sorry @cuongtm95 the merge was meant to be as soon as Attack_system is merged into master.. it contains the logic and methods that I talked about in my previous comment. I'm already trying to reach @Kaes3kuch3n to approve it.
As soon as it is merged, please merge again. I will do a review afterwards.

SamuelllG:
Merged now!

cuongtm95:
> Merged now!

Yeah was just cleaning up the code and fix some conflicts. Will do it tonight when I get home.

cuongtm95:
The feature is now fully working ! Waiting on review..

Assets/Scripts/Interactables/Crate.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.Collections.Generic;
+using UnityEngine;
+
+public class Crate : Interactable
+{
+    Animator animator;
+
+    [SerializeField] bool isOpen;
+
+    private TooltipDisplay tooltipdisplay;
```
Kaes3kuch3n: Looks like this isn't used

Assets/Scripts/Interactables/Interactable.cs
```diff
@@ -0,0 +1,71 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class Interactable : MonoBehaviour
+{
+    [Tooltip("The radius around the object that stops character going inside it")]
+    [SerializeField] public float radius = 3f;
```
Kaes3kuch3n: If you use `[SerializeField]`, you don't have to define the field as public

Assets/Scripts/Interactables/Interactable.cs
```diff
@@ -0,0 +1,71 @@
...
+
+    [Tooltip("Is this interactable currently being focused?")]
+    bool isFocus = false;
```
Kaes3kuch3n: Please add access modifiers for all fields :) (`isFocus`, `hasInteracted` and `player`)

Assets/Scripts/Interactables/Interactable.cs
```diff
@@ -0,0 +1,71 @@
...
+
+    [Tooltip("Have we already interacted with the object?")]
+    bool hasInteracted = false;
+
+    Transform player;    // Reference to the player transform
+
+
+    void Update()
```
Kaes3kuch3n: Please also add an access modifier here :) Should be private, I guess

Assets/Scripts/Interactables/Interactable.cs
```diff
@@ -0,0 +1,71 @@
...
+        isFocus = false;
+        hasInteracted = false;
+        player = null;
+    }
+
+    /// <summary>
+    /// This method is meant to be overwritten for each interactable objects 
+    /// (for example: OPEN if it was a "crate" or CHOP if it was a "tree"...)
+    /// </summary>
+    public virtual void Interact()
```
Kaes3kuch3n: Since this method isn't implemented in here, it would make more sense to make this class and this method abstract, so that overriding it is obligatory

Assets/Scripts/Interactables/ItemPickup.cs
```diff
@@ -0,0 +1,37 @@
...
+public class ItemPickup : Interactable
+{
+    [SerializeField] public Item item; //reference for items
+
+    [SerializeField] public Inventory inventory; // reference for the inventory
+
+    // Unity Events
+    private void Start()
+    {
+        inventory = GameObject.Find("Player").GetComponentInChildren<Inventory>(); //auto assign the Inventory to the component
```
Kaes3kuch3n: You can use the player manager to get the player reference, and you can access the inventory field of the player if you create a getter for it :) And the inventory field doesn't have to be public if you get the reference yourself

Assets/Scripts/Player/PlayerController.cs
```diff
@@ -111,8 +118,63 @@ private void Update()
...
+        // This will make character moving to the target object
+        public void MovingToTarget(Interactable newTarget)
+        {
+            _navMeshAgent.stoppingDistance = newTarget.radius * .8f;
+            _navMeshAgent.updateRotation = false;
+            target = newTarget.transform;
+        }
+
+        // Makes character stop moving to target object
+        public void StopMovingToTarget()
```
Kaes3kuch3n: The nav mesh agent should stop automatically if it's close enough to the target. Changing the stopping distance manually isn't the best idea.

Assets/Scripts/Player/PlayerController.cs
```diff
@@ -111,8 +118,63 @@ private void Update()
...
+        }
+
+        /// <summary>
+        /// These are functions for character to interact with all objects in the game (ex: moving, following, pick up items, open crate, etc..) using raycast and _NavMeshAgent
+        /// (Won't conflict with character attacking enemy or moving on the map)
+        /// </summary>
+        #region Interactables Functions
+        
+        // This will make character moving to the target object
+        public void MovingToTarget(Interactable newTarget)
```
Kaes3kuch3n: I think this is something that can be combined with the regular movement logic, since it's basically the same

Assets/Scripts/Player/PlayerController.cs
```diff
@@ -111,8 +118,63 @@ private void Update()
...
+        // Remove focus when not interact
+        public void RemoveFocus()
+        {
+            if (focus != null) focus.OnDefocused();
+            focus = null;
+            StopMovingToTarget();
+        }
+
+        // Makes character facing the target when moving toward it
+        void FaceTarget()
```
Kaes3kuch3n: We already have a FaceTarget method. I think you could also just use the existing one :)

Assets/Scripts/UI/TooltipDisplay.cs
```diff
@@ -0,0 +1,56 @@
...
+    [Tooltip("Reference to Text UI element")]
+    [SerializeField] public Text myText;
+
+    [Tooltip("Determine if we want to display the UI or not")]
+    [SerializeField] public bool displayInfo;
+
+    private void Start()
+    {
+        myText = gameObject.GetComponentInChildren<Text>();
+        tooltip = gameObject.GetComponentInChildren<Canvas>();
```
Kaes3kuch3n: GetComponentInChildren is not really performant. Since the fields for this are public, it would be better if you would just reference them in the inspector

Assets/Scripts/UI/TooltipDisplay.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+        if (displayInfo)
+        {
+            myText.text = myString;
+            tooltip.enabled = true;
+        }
+        else
+        {
+            tooltip.enabled = false;
+        }
```
Kaes3kuch3n: I think it would be better to move this logic to `OnMouseOver` and `OnMouseExit`, since otherwise it is called every time in `Update()`

Kaes3kuch3n (changes requested):

Assets/Scripts/Player/PlayerController.cs
cuongtm95: This is a "naming" mistake from me. Actually this will make the character not targeting the object anymore and change the stopping distance to 0 so the character can move all the way to where the mouse click without stopping at a distance.

Assets/Scripts/Player/PlayerController.cs
cuongtm95: Same as above, this should be called SetTarget instead of MovingToTarget. Sorry. This will make the character stop when it is close enough to the target by setting the stopping distance = the "radius" value of the object (can be tweaked)

Assets/Scripts/Interactables/Crate.cs
cuongtm95: Oh yeah, already deleted !

Assets/Scripts/Interactables/Interactable.cs
cuongtm95: I have to define it as public so i can access the value in PlayerController.

Assets/Scripts/Interactables/Interactable.cs
cuongtm95: done

Assets/Scripts/Interactables/Interactable.cs
cuongtm95: I think you're right. At first i want to have something in the interact method here like "play sound, or visual or something" but maybe i will just make it abstract.

Assets/Scripts/Interactables/ItemPickup.cs
cuongtm95: oh ya. i forgot about PlayerManager. Already done.

Assets/Scripts/UI/TooltipDisplay.cs
cuongtm95: Yeah i'm consider changing the whole class and methods for Tooltip. Will update soon.

Assets/Scripts/Interactables/Interactable.cs
Kaes3kuch3n: If that's the case, you don't have to add the [SerializeField] attribute. But since I guess the player controller shouldn't be able to change the radius, it would be even better if you created a getter for the field and make it private:
```c#
public float Radius => radius;
```

Assets/Scripts/Interactables/Interactable.cs
Kaes3kuch3n: Yeah, I think it can be abstract for now. If we want to add some general implementation later on, we can still change it back 👍

Kaes3kuch3n (changes requested):

Assets/Scripts/Player/PlayerController.cs
cuongtm95: Okay !

Assets/Scripts/UI/TooltipDisplay.cs
cuongtm95: For now i will remove the tooltip script and stuffs. Will make a new PR for that.

Assets/Scripts/Interactables/Interactable.cs
cuongtm95: done

cuongtm95:
For now i will remove the tooltip script and stuffs. Will make a new PR for that.
Player now can use Right Click to pick up items and it will be automatically added to Inventory.
Player now can use Right Click to interact with any Object that had Interactable.cs attached to it (i added Tree.cs for later purpose), for example: axe, crate, tree, enemies.. 
We can use this to implement many things in the future by overriding **Interact()** since it's a abstract method, like "play Sound" or "play animation", etc..)

Assets/Scripts/Interactables/Crate.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class Crate : Interactable
+{
+    Animator animator;
+
```
SamuelllG, line 8: access modifiers

Assets/Scripts/Interactables/Crate.cs
```diff
@@ -0,0 +1,24 @@
...
+    [SerializeField] bool isOpen;
+
+    void Start()
+    {
+        animator = GetComponent<Animator>();
```
SamuelllG: As far as I learned from @DavidPgl it should be in Awake or maybe a SerializeField?

Assets/Scripts/Interactables/Interactable.cs
```diff
@@ -0,0 +1,70 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public abstract class Interactable : MonoBehaviour
+{
+    [Tooltip("The radius around the object that stops character going inside it")]
+    private float radius = 3f;
```
SamuelllG: please don't put values here

Assets/Scripts/Managers/PlayerManager.cs
```diff
@@ -5,13 +5,18 @@ namespace Managers
...
         public GameObject GetPlayer()
         {
             if (player == null)
                 player = GameObject.FindGameObjectWithTag("Player"); //BUG: Replace me with Trigger
             return player;
         }
 
+        public Inventory GetInventory()
+        {
+            return inventory = player.GetComponentInChildren<Inventory>();
```
SamuelllG: why make it SerializeField if you use GetCompInChildren?

Shouldn't the PlayerManager only have the player anyway? I think you should get the inventory via the player that you get from the playermanager. Correct me, if I'm wrong

Assets/Scripts/Player/PlayerController.cs
```diff
@@ -160,6 +218,17 @@ private void OnRightClick(InputAction.CallbackContext obj)
...
                     Move(groundHit);
                 }
+        
+                if (interactable != null)
+                {
+                    //Set as focus
+                    SetFocus(interactable);
+                }
+
+                else
```
SamuelllG: you should also defocus when a Damageable is clicked

Assets/Scripts/Player/PlayerController.cs
```diff
@@ -111,7 +120,54 @@ private void Update()
...
+
+            // Check and makes character facing the target object when moving toward it
+            if (target != null) FaceTarget(target.gameObject, true, out _);
+        }
+
+        /// <summary>
+        /// These are functions for character to interact with all objects in the game (ex: moving, following, pick up items, open crate, etc..) using raycast and _NavMeshAgent
+        /// (Won't conflict with character attacking enemy or moving on the map)
+        /// </summary>
+        #region Interactables Functions
```
SamuelllG: pls put all of this in a class InteractLogic

Assets/Scripts/Player/PlayerController.cs
```diff
@@ -111,7 +120,54 @@ private void Update()
...
+        
+        // This will make character going to target object and stop when it is close enough
+        public void SetTarget(Interactable newTarget)
```
SamuelllG: could you make the methods private?

Assets/Scripts/Player/PlayerController.cs
```diff
@@ -111,7 +120,54 @@ private void Update()
...
+        {
+            _navMeshAgent.stoppingDistance = newTarget.Radius * .8f;
+            _navMeshAgent.updateRotation = false;
+            target = newTarget.transform;
+        }
+
+        // Makes character not targeting the object anymore
+        public void StopTarget()
+        {
+            _navMeshAgent.stoppingDistance = 0f;
```
SamuelllG: you set the stoppingDistance to 0 😮 ?

Assets/Scripts/Player/PlayerController.cs
```diff
@@ -160,6 +218,17 @@ private void OnRightClick(InputAction.CallbackContext obj)
                     _attackLogic.StopAttack();
                     Move(groundHit);
                 }
+
```
SamuelllG: that's kinda unclear. If there is no damageable, then move, after that you check for an interactable (when it's already moving). It should be if there is no damageable, check for an interactable, if there is no interactable, then move. You should include moving towards the interactable in your InteractLogic class like done for attacking.
I suggest something like:
```
Interactable interactable = objectHit.GetComponent<Interactable>();
IDamageable damageable = objectHit.GetComponent<IDamageable>();

if (damageable != null)
{
     _interactLogic.RemoveFocus();
     _attackLogic.StartAttack(objectHit);
}
else if (interactable != null)
{
     _attackLogic.StopAttack();
     _interactLogic.SetFocus(interactable);
}
else if (Physics.Raycast(clickRay, out RaycastHit groundHit, 10000, moveClickLayers))
{
     _attackLogic.StopAttack();
     _interactLogic.RemoveFocus();
     Move(groundHit);
}

```

Assets/Scripts/Player/PlayerController.cs
```diff
@@ -111,7 +120,54 @@ private void Update()
             Vector3 playerPosition = transform.position;
             _camera.transform.position = playerPosition + _cameraPosition;
             _camera.transform.LookAt(playerPosition);
+
+            // Check and makes character facing the target object when moving toward it
+            if (target != null) FaceTarget(target.gameObject, true, out _);
```
SamuelllG: why do you set updateRotation to false below? This line is just needed because of that.

SamuelllG: with implementation of an InteractLogic class not needed anymore

Assets/Scripts/Interactables/ItemPickup.cs
```diff
@@ -0,0 +1,29 @@
...
+        PickUp();
+    }
+
+    /// <summary>
+    /// Function for picking up item, add to inventory and then make item disappeared
+    /// </summary>
+    public void PickUp()
+    {
+        //Add to inventory
+        PlayerManager.Instance.GetInventory().Add(item);
```
SamuelllG: Can an item be picked up multiple times, for example after throwing it away from inventory? Then I think you should get the inventory once, and then always add to it. Don't get it via the playermanager > player every time the item is picked up.

SamuelllG (changes requested):
I suggest implementing an InteractLogic class :)

Assets/Scripts/Interactables/Crate.cs
cuongtm95: is done !

Assets/Scripts/Interactables/Crate.cs
cuongtm95: Alright ! changed it to Awake()

Assets/Scripts/Interactables/Interactable.cs
cuongtm95: I just want to set a default value for it. Will do it in Start() then.

Assets/Scripts/Managers/PlayerManager.cs
cuongtm95: I'm not sure if PlayerManager should only have the player. If so i will create a getter for inventory in other place.

Assets/Scripts/Player/PlayerController.cs
cuongtm95: Oki !

Assets/Scripts/Player/PlayerController.cs
cuongtm95: is done

Assets/Scripts/Player/PlayerController.cs
cuongtm95: this looks good to me.

Assets/Scripts/Player/PlayerController.cs
cuongtm95: This basically mean that if the character not targeting anything then it will be switched back to normal, that means the character will not stop before reaching the position where the mouse clicked.

Assets/Scripts/Player/PlayerController.cs
cuongtm95: Yeah just removed updateRotation. Seems like it is not needed

Assets/Scripts/Interactables/ItemPickup.cs
cuongtm95: Yeah changed to this 
```suggestion
        private void Awake()
    {
        GetInventory();
    }

    private Inventory GetInventory()
    {
        return inventory = PlayerManager.Instance.GetPlayer().GetComponentInChildren<Inventory>();
    }
```
should be working fine

cuongtm95: For now since we are not able to drop the item yet so i can't test it but im pretty sure it will work just fine.

cuongtm95:
- Moved everything Interact-related from PlayerController to InteractLogic.cs and cleaned some unneeded code.
- Get Inventory once on ItemPickup at Awake()

Assets/Scripts/InteractLogic.cs
```diff
@@ -0,0 +1,79 @@
...
+
+    [Tooltip("Showing the object player is focusing")]
+    [SerializeField]
+    private Interactable focus;
+
+    [Tooltip("Showing the object target player is interacting")]
+    [SerializeField]
+    private Transform target;
+
+    private IMovable _movable;
```
SamuelllG: We don't have IMovable anymore, if you pull master to resolve merge conflicts, you'll see. Please resolve merge conflicts

Assets/Scripts/Interactables/Interactable.cs
SamuelllG: thanks!

Assets/Scripts/Managers/PlayerManager.cs
SamuelllG: @tobihoffmann @DavidPgl @Kaes3kuch3n @potatoTVnet @thuemmels

Assets/Scripts/Player/PlayerController.cs
SamuelllG: You're right, default stoppingDistance is 0. I confused things here!

Assets/Scripts/Interactables/ItemPickup.cs
SamuelllG: thanks!

Assets/Scripts/InteractLogic.cs
```diff
@@ -0,0 +1,79 @@
...
+
+    private NavMeshAgent _navMeshAgent;
+
+    private void Awake()
+    {
+        _movable = GetComponent<IMovable>();
+        _navMeshAgent = GetComponent<NavMeshAgent>();
```
SamuelllG: You could think about giving Movable a public method SetStoppingDistance. In my opinion it's better to do it all via the Movable instead of getting the NavMeshAgent here.

Assets/Scripts/InteractLogic.cs
```diff
@@ -0,0 +1,79 @@
...
+
+    private void Update()
+    {
+        // Check and makes character facing the target object when moving toward it
+        if (target != null) _movable.FaceTarget(target.gameObject, true, out _);
+    }
+    // This will make character going to target object and stop when it is close enough
+    private void SetTarget(Interactable newTarget)
+    {
+        _navMeshAgent.stoppingDistance = newTarget.Radius * .8f;
```
SamuelllG: Is there a reason for * .8f ?

Assets/Scripts/InteractLogic.cs
```diff
@@ -0,0 +1,79 @@
...
+    /// </summary>
+    /// 
+
+    [Tooltip("Showing the object player is focusing")]
+    [SerializeField]
+    private Interactable focus;
+
+    [Tooltip("Showing the object target player is interacting")]
+    [SerializeField]
+    private Transform target;
```
SamuelllG: Should focus and target be SerializeFields? Are they meant to be set via inspector?

Assets/Scripts/Player/PlayerController.cs
```diff
@@ -147,19 +153,32 @@ private void OnRightClick(InputAction.CallbackContext obj)
             if (Physics.Raycast(clickRay, out RaycastHit hit, 10000) && _attackLogic.Status == AttackLogic.AttackStatus.None)
             {
                 GameObject objectHit = hit.collider.gameObject;
+                Interactable interactable = hit.collider.GetComponent<Interactable>();
                 IDamageable damageable = objectHit.GetComponent<IDamageable>();
+
                 if (damageable != null)
                 {
                     // implementation NOT capable of area damage
```
SamuelllG: Don't forget to RemoveFocus here :) or was it on purpose?

Assets/Scripts/Player/PlayerController.cs
```diff
@@ -147,19 +153,32 @@ private void OnRightClick(InputAction.CallbackContext obj)
             if (Physics.Raycast(clickRay, out RaycastHit hit, 10000) && _attackLogic.Status == AttackLogic.AttackStatus.None)
             {
                 GameObject objectHit = hit.collider.gameObject;
+                Interactable interactable = hit.collider.GetComponent<Interactable>();
```
SamuelllG: objectHit.GetComp
It's just a minor thing, but it makes it a little bit faster I guess :]

Assets/Scripts/Player/PlayerController.cs
SamuelllG: Thanks for implementing! with interactlogic it looks much better 👍

SamuelllG (changes requested):
I commented a few things. Please resolve merge conflicts :)

Assets/Scripts/InteractLogic.cs
cuongtm95: Just for us to see via inspector what the object is. Nothing need to be set here.

Assets/Scripts/InteractLogic.cs
cuongtm95: No actually. I just hard coded it and forgot to change back.

Assets/Scripts/Player/PlayerController.cs
cuongtm95: Oh yeah i want to also focus on the damageable as well, so addition to attack we can do something more if we attach Interactable.cs on it. Without Interactable.cs nothing will be focused.

Assets/Scripts/Player/PlayerController.cs
cuongtm95: gotcha

Assets/Scripts/InteractLogic.cs
cuongtm95: is done !
```suggestion
        public void SetStoppingDistance(float distance)
        {
            NavMeshAgent.stoppingDistance = distance;
        }
```

cuongtm95:
```suggestion
        public void SetStoppingDistance(float distance)
        {
            NavMeshAgent.stoppingDistance = distance;
        }
```

cuongtm95: is done !
```suggestion
        public void SetStoppingDistance(float distance)
        {
            NavMeshAgent.stoppingDistance = distance;
        }
```

```suggestion
       _movable.SetStoppingDistance(newTarget.Radius);
```

Assets/Scripts/Entity/Player/PlayerController.cs
```diff
@@ -148,19 +154,31 @@ private void OnRightClick(InputAction.CallbackContext obj)
...
                 GameObject objectHit = hit.collider.gameObject;
+                Interactable interactable = objectHit.GetComponent<Interactable>();
                 Damageable damageable = objectHit.GetComponent<Damageable>();
                 if (damageable != null)
                 {
                     // implementation NOT capable of area damage
                     _attackLogic.StartAttack(objectHit);
                 }
+
+                else if (interactable != null)
```
Kaes3kuch3n:
```suggestion
                else if (objectHit.GetComponent<Interactable>() is Interactable interactable)
```
I'd rewrite this a little so that the variable declaration and the null check are joined together, because we don't need the variables outside of this scope, so this would be a little clearer :) And if you're already on it, can you change the damageable check the same way? 😅

Assets/Scripts/InteractLogic.cs
Kaes3kuch3n: If you just need to see them in the inspector, you can also remove the SerializeField attributes and simply switch the inspector to debug mode. In this mode it also displays the private fields that don't have the attribute

Assets/Scripts/Interactables/ItemPickup.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        [SerializeField] private Item item; //reference for items
+        [SerializeField] private InventoryController inventory;
+
+        private void Start()
+        {
+            GetInventory();
+        }
+
+        private InventoryController GetInventory()
```
Kaes3kuch3n: I think this isn't really needed, since you can simply drag in the reference to the inventory into the serialised field :)

Kaes3kuch3n (changes requested):
Almost there, just a few small change requests left :)

Assets/Scripts/Interactables/ItemPickup.cs
cuongtm95: Oh okay. I thought we don't want to manually assign anything so...

Assets/Scripts/Entity/Player/PlayerController.cs
cuongtm95: Alright !

cuongtm95:
Should be okay now.

cuongtm95:
Should be fine now. Please check again and merge to master without GameScene

Kaes3kuch3n (dismissed):

Assets/Scripts/Player/PlayerController.cs
SamuelllG: I think you misunderstood me, @Kaes3kuch3n updated the code

SamuelllG (dismissed):

inkompotato (approved):
tested by Luis - trusted by me

SamuelllG (approved):

PR #30, opened 2019-12-03
Npc's find near enemies

DavidPgl:


Assets/Scripts/Enemy/EnemyController.cs
```diff
@@ -0,0 +1,234 @@
...
+        private float rotationTolerance;
+        
+        [Tooltip("The radius around the NPC in which a player can escape.")] [SerializeField]
+        private float escapeRadius;
+
+        [Tooltip("Show escape radius in scene view.")] [SerializeField]
+        private bool showEscapeRadius;
+
+        [Tooltip("TargetFinderComponent - Finds nearby targets")] [SerializeField]
+        private TargetFinder targetFinder;
```
DavidPgl: New

Assets/Scripts/Enemy/EnemyController.cs
```diff
@@ -0,0 +1,234 @@
...
+        {
+            if (!_isChasing)
+                Wander();
+
+            if (!isAggressive || targetFinder.Targets.Count == 0) return;
+
+            // Attack first target in list
+            GameObject currentTarget = targetFinder.Targets[0];
+
+            if (!_isChasing)
```
SamuelllG:
```
if (!isAggressive || targetFinder.Targets.Count == 0) Wander();

// Attack first target in list
GameObject currentTarget = targetFinder.Targets[0];

if (!_isChasing) StartChasing();
```

Assets/Scripts/Enemy/TargetFinder.cs
```diff
@@ -0,0 +1,32 @@
...
+    [RequireComponent(typeof(Collider))]
+    public class TargetFinder : MonoBehaviour
+    {
+        [Tooltip("Every layer that will be considered as possible target")] [SerializeField]
+        private LayerMask attackableLayers;
+
+        /// <summary>
+        /// List of all targets that entered the trigger-radius and have a layer of the <see cref="attackableLayers"/> layer mask
+        /// </summary>
+        public readonly List<GameObject> Targets = new List<GameObject>();  //TODO: Should probably be replaced by the abstract class "Damageable"
```
SamuelllG: Concerning the TODO: Like I said, AttackLogic needs the target's position, so passing Damageable won't be enough.

Assets/Scripts/Enemy/TargetFinder.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        private void OnTriggerEnter(Collider other)
```
SamuelllG: Where do you define the radius of the trigger? It should be lookRadius.

Assets/Scripts/Utils/ClassExtender.cs
```diff
@@ -0,0 +1,15 @@
+﻿using UnityEngine;
+
+namespace Utils
+{
+    public static class ClassExtender
```
SamuelllG, line 5: Where is this even used?

SamuelllG (changes requested):

SamuelllG:
When checking out your branch, I get several NullRefs and can't move! Additional AttackLogic scripts in prefabs all are resetted

Assets/Scripts/Enemy/TargetFinder.cs
SamuelllG: Oh okay you have an additional gameObject with a trigger collider. It's a piti that you can't change the attackRadius from the enemy object anymore. At least you see the targetFinder SerializedField there.

DavidPgl: Physics.Raycast and colliders made me some porblems. That was the only way I could think of atm

Assets/Scripts/Utils/ClassExtender.cs
DavidPgl: ` if(!attackableLayers.Contains(other.gameObject.layer)) return;`
See targetFinder line 25

Assets/Scripts/Enemy/TargetFinder.cs
SamuelllG: Yep, I think it's okay 👍

Assets/Scripts/Enemy/TargetFinder.cs
DavidPgl: The abstract class should extend from MonoBehaviour or be an component for itself maybe?

Assets/Scripts/Enemy/EnemyController.cs
DavidPgl: This would prevent an aggressive enemy from wandering, right?

DavidPgl:
^This should fix the nullrefs when going back to the main menu and starting the game again

Assets/Scripts/Enemy/EnemyController.cs
Kaes3kuch3n: I think this has to be reworked later to support passive, neutral and aggressive NPCs

Assets/Scripts/Enemy/TargetFinder.cs
Kaes3kuch3n: How about adding something like GetPosition() to Damageable?

Kaes3kuch3n (dismissed):

Kaes3kuch3n (approved):

inkompotato (approved):
cool

SamuelllG (approved):

PR #53, opened 2019-12-07
Player can use items

thuemmels:
Things to improve: 
(I'd rather have this stuff merged with master first so everyone can work with the same stuff)

- how to handle resources such as wood, rocks that are neither equipable nor consumable? maybe new class for them?
- implement max stack size 
- code duplication in consumable and equipable: both implement Equals in the same way, but I can't move it up in the hierarchy because Item is an abstract class.
- heal "animation" is the same as getting hit, because it uses the same function right now


DavidPgl:
What about:
- Everything is an Item
- Consumable and/or equipable items will be extended with the needed behaviour
This makes it easier for wood and stone and keeps them as collectible
(So items is not abstract anymore)
But this depends on wood and stone not having anything special that consumable or equibable item don't have

Assets/Scripts/Inventory/Consumable.cs
```diff
@@ -0,0 +1,36 @@
...
+    private int saturationValue;
+    [SerializeField]
+    private int hydrationValue;
+
+    /// <summary>
+    /// Gets the PlayerState of the player, and changes its values depending on the item values.
+    /// </summary>
+    public override bool Use() 
+    {
+        PlayerState playerState = Singleton<PlayerManager>.Instance.GetPlayer().GetComponent<PlayerState>();
```
DavidPgl: Two things:
You can use `PlayerManager.Instance.GetPlayer()`
Also, maybe it's better to give the PlayerManager this functionality, so you just call PlayerManager.Instance.Change... ?
Not sure how good that is but I don't like this long line haha

DavidPgl (dismissed):

Assets/Scripts/Inventory/Consumable.cs
```diff
@@ -0,0 +1,36 @@
...
+        if (healthValue > 0)
+            playerState.Hit(-healthValue);
```
SamuelllG: you use the Hit method with a negative damage value to heal the player? 👀 😂 why not just use ChangePlayerHealth?

SamuelllG (changes requested):
I like David's idea :)

Assets/Scripts/Inventory/Consumable.cs
Kaes3kuch3n: Currently `ChangePlayerHealth` is private, so we would either have to change that or add another public method `void Heal(int amount)` or something like that

tobihoffmann: or - if it needs more than the change of healthValue in the future (like triggering an animation or sound in some way) - implement a new medthod to playerState?

edit:
posted at the same time as Luis ... Lets do this

SamuelllG: > 
> 
> Currently `ChangePlayerHealth` is private, so we would either have to change that or add another public method `void Heal(int amount)` or something like that

as we had discussed already, yep, I like this 👍

Assets/Scripts/Inventory/Consumable.cs
thuemmels: I shortened the line, but decided to skip the refactoring for now.

Assets/Scripts/UI/ItemButton.cs
```diff
@@ -27,5 +59,24 @@ public int Count
...
             }
         }
+
+        /// <summary>
+        /// If the button is clicked, the item saved in the _item field is used. Consumables are going to be removed
+        /// from the inventory.
+        /// </summary>
+        public void OnClick() 
+        {
+            if (_item is Consumable)
```
Kaes3kuch3n:
```suggestion
            if (_item is Consumable consumable)
```
Dann kannst Du unten gleich consumable.Use() schreiben und musst nicht noch extra casten :) (Und für Equipable auch)

Assets/Scripts/UI/ItemButton.cs
```diff
@@ -14,6 +15,37 @@ public class ItemButton : MonoBehaviour
...
+
+        private Button _button;
+
+        private void Awake()
+        {
+            _button = GetComponent<Button>();
+            _button.onClick.AddListener(OnClick);
+        }
+
+        public Item Item
```
Kaes3kuch3n: Können die dann nicht direkt public fields sein? Können ja auch [HideInInspector] sein

Kaes3kuch3n (changes requested):
Nur ne Kleinigkeit :)

Kaes3kuch3n (dismissed):

Assets/Scripts/Inventory/Consumable.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// <summary>
+    /// Gets the PlayerState of the player, and changes its values depending on the item values.
+    /// </summary>
+    public new bool Use() 
+    {
+        PlayerState playerState = PlayerManager.Instance.GetPlayer().GetComponent<PlayerState>();
+        if (healthValue > 0)
+            playerState.Heal(healthValue);
+        if (saturationValue > 0)
+            playerState.ChangePlayerSaturation(saturationValue);
```
SamuelllG: Implement Consume() as discussed and make ChangePlayer___ private

SamuelllG (changes requested):

Kaes3kuch3n (dismissed):

SamuelllG (dismissed):

Kaes3kuch3n (dismissed):

Assets/Scripts/UI/ItemButton.cs
```diff
@@ -14,6 +15,20 @@ public class ItemButton : MonoBehaviour
...
+        public Item _item;
+
+        [HideInInspector]
+        public InventoryController _inventory;
+
+        private Button _button;
+
+        private void Awake()
+        {
+            _button = GetComponent<Button>();
```
DavidPgl, line 29: I think atm there is no button on the ItemButton Prefab. It could definitely cause some problems with the dragging and sorting of items. If that's the case, let me know:)

DavidPgl (changes requested):

Assets/Scripts/UI/ItemButton.cs
DavidPgl: I can have a look into this later today

Kaes3kuch3n:
Fixed the merge conflicts and added the missing button component. Should be working fine now 👍

Kaes3kuch3n (approved):
😏

DavidPgl (approved):

PR #55, opened 2019-12-09
Sanity system: Stat + basic influence

SamuelllG:
This is a first part of the sanity system. It includes: New stat Sanity influenced by other Stats: Health, Hunger, Thirst.

Tweaking will take some time, I just put values that were kinda acceptable :)


Assets/Scripts/Utils/Enums.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Utils
+{
+    public class Enums : MonoBehaviour
```
DavidPgl: We should have different classes for each type of enum. Probably better to make it directly be enum StatType

Assets/Scripts/Entity/Player/Sanity/SanityMath.cs
```diff
@@ -0,0 +1,108 @@
...
+        /// <param name="value">Current value of stat</param>
+        public void InfluenceSanityByStat(StatType statType, int value)
+        {
+            Stat stat = GetStatByType(statType);
+            if (value < stat.Boundary)
+                stat.Rate = (value - stat.Boundary) * stat.Severity;
+            else
+                // don't use - stat.Boundary because then if you have 100% of stat,
+                // you get a higher positive rate, the lower the boundary
+                stat.Rate = value * gainFactor * stat.Severity;
```
DavidPgl, line 57: Probably more of a implementation question:
Should saturation and hydration influence the sanity positively? Or should only actions (e.g. sleeping) and/or using specific items (e.g. cooked food or better food) influence it that way?

Assets/Scripts/Utils/Stat.cs
```diff
@@ -0,0 +1,26 @@
...
+        [Tooltip("The stat boundary for sanity impact. At this value, the stat doesn't influence sanity (neither positively nor negatively)")]
+        [SerializeField]
+        public int Boundary;
+
+        [Tooltip("The severity at which the stat influences sanity")]
+        [SerializeField]
+        public float Severity;
+
+        // the rate at which the stat influences sanity
+        public float Rate { get; set; }
```
DavidPgl: Specific reason for property here?

Assets/Scripts/Entity/Player/Sanity/SanityMath.cs
```diff
@@ -0,0 +1,108 @@
...
+        /// ones with a lower severity. 
+        /// </summary>
+        private void UpdateChangeRate()
+        {
+            bool allPositive = AllRatesPositive();
+            _changeRate = 0;
+            foreach (Stat stat in stats)
+            {
+                // only use negative rates or all if no stat is negative
+                if (!allPositive && stat.Rate < 0 || allPositive)
```
Kaes3kuch3n:
```suggestion
                if (allPositive || stat.Rate < 0)
```

Assets/Scripts/Entity/Player/Sanity/SanityMath.cs
```diff
@@ -0,0 +1,108 @@
...
+                    // found a positive rate between negative rates
+                    // --> use it, but with lower impact
+                    _changeRate += stat.Rate * lowImpactSeverity;
+            }
+        }
+
+        /// <summary>
+        /// Gets the stat from <see cref="stats"/>
+        /// </summary>
+        private Stat GetStatByType(StatType statType)
```
Kaes3kuch3n: Would change this to
```c#
stats.Where(s => s.Type == statType).FirstOrDefault();
```
Then it returns null instead of throwing an exception, which makes it easier to handle the case that a stat is not found

Assets/Scripts/Entity/Player/Sanity/SanityMath.cs
```diff
@@ -0,0 +1,108 @@
...
+                if (stat.Type == statType)
+                    return stat;
+            }
+            throw new Exception("Stat not found");
+        }
+
+        /// <summary>
+        /// Checks if all stat rates are positive
+        /// </summary>
+        private bool AllRatesPositive()
```
Kaes3kuch3n, line 93: You could also change this to
```c#
return stats.All(s => s.Rate >= 0)
```

tobihoffmann (commented):
Shouldnt sanity be a property of playerState and let SanityController influence it (for clarity)?
Might also be nice for syncing values in Multiplayer

Assets/Scripts/Entity/Player/Sanity/SanityController.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        /// <summary>
+        /// Sums up a tick weighting the total change rate with <see cref="severity"/>. 
+        /// Only sums up a positive tick if player doesn't have full sanity. Ticks are either
+        /// positive or negative 1. If summed up tick, changes player sanity.
+        /// </summary>
+        private void TickSanity()
+        {
+            _nextTick += sanityMath.GetCurrentChange();
+            if (sanity == 100 && _nextTick > 0)
```
Kaes3kuch3n: Are you only checking the upper bound on purpose, or did you forget to check the lower bound when decreasing the sanity? 😅

Kaes3kuch3n (changes requested):
Looks pretty good already, I'd say 👍

Assets/Scripts/Entity/Player/Sanity/SanityMath.cs
Kaes3kuch3n: I think it's okay for it to be influenced by sufficient saturation/hydration, too. As long as the boundary isn't too low, that is :D

Assets/Scripts/Utils/Enums.cs
SamuelllG: I don't get it, what do you suggest exactly?

Assets/Scripts/Entity/Player/Sanity/SanityMath.cs
SamuelllG: Yep I agree

Assets/Scripts/Utils/Stat.cs
SamuelllG: eh, no :D

SamuelllG:
> 
> 
> Shouldnt sanity be a property of playerState and let SanityController influence it (for clarity)?
> Might also be nice for syncing values in Multiplayer

That's what I had at first. I put it in the sanity controller because I didn't want to much interaction going on there, but the way I finally implemented it could have the sanity stat in the playerstate very easily. I just didn't move it back. If you prefer having it in the player state, I'll move it back.

Assets/Scripts/Entity/Player/Sanity/SanityMath.cs
SamuelllG: That's what I missed, thanks!

SamuelllG: "using System.Linq" 😮 how should I've found this?

Assets/Scripts/Entity/Player/Sanity/SanityMath.cs
SamuelllG: so pretty :)

Assets/Scripts/Entity/Player/Sanity/SanityController.cs
SamuelllG: I check upper bound only, because I assumed the player dies when sanity goes 0. I don't know if we want that, but I'd prefer it because it really motivates you to do something when you notice mobs are spawning around you :)
Independently of whether we want that or not, I think building up a "negative malus" when sanity = 0 for the lower bound is alright as well, don't you think?

SamuelllG:
@DavidPgl your review didn't count as a review? :(

Assets/Scripts/Utils/Enums.cs
DavidPgl: Don't use an Enum-Class, make an enum directly:
`public enum StatTypes`
And I would probably make this an inner enum as well. It is only related to the stats and will never be used in other context.

tobihoffmann:
> > Shouldnt sanity be a property of playerState and let SanityController influence it (for clarity)?
> > Might also be nice for syncing values in Multiplayer
> 
> That's what I had at first. I put it in the sanity controller because I didn't want to much interaction going on there, but the way I finally implemented it could have the sanity stat in the playerstate very easily. I just didn't move it back. If you prefer having it in the player state, I'll move it back.

I'd prefer it to be in PlayerState just to have all statuses grouped together for a consistent structure. As in if i wouldn't know where to look for the property my first guess would be there. 
But this absolutely up for discussion.

Assets/Scripts/Entity/Player/Sanity/SanityController.cs
Kaes3kuch3n: Yeah, it's fine for now 👍

Kaes3kuch3n (commented):
Just fix David's enum request, then I'll approve it :P

Kaes3kuch3n (dismissed):

Assets/Scripts/Utils/Enums.cs
SamuelllG: okay I put it in SanityController

SamuelllG:
I implemented the enum as suggested. Also did put the sanity in PlayerState, so we have a bit more of coupling going on.. also had to make the ChangePlayerSanity method public not like the other ones, but I really couldn't think of a method "makeInsane" xD ... so idk about consistency but anyways

(sorry for the double commit btw)

Kaes3kuch3n (approved):

DavidPgl (approved):

PR #56, opened 2019-12-09
Crafting logic

DavidPgl:


Assets/Scripts/AbstractClasses/IItemHandler.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// <param name="amount">The amount of this item to be present</param>
+        /// <returns>Whether the item and the given amount is present</returns>
+        bool ContainsItem(Item item, int amount = 1);
+
+        /// <summary>
+        /// Adds an item with a specified quantity
+        /// </summary>
+        /// <param name="item">The item to be added</param>
+        /// <param name="amount">The amount to be added</param>
+        void AddItem(Item item, int amount);
```
SamuelllG: sounds like a hack ;)

Assets/Scripts/Crafting/CraftingRecipe.cs
```diff
@@ -0,0 +1,71 @@
...
+            foreach (ResourceData createdItem in createdItems)
+                itemHandler.AddItem(createdItem.item, createdItem.amount);
+        }
+
+
+        /// <summary>
+        /// Small struct for storing the item and the amount of the item as an "ingredient" for the recipe.
+        /// </summary>
+        [Serializable]
+        public struct ResourceData
```
SamuelllG, line 65: Shouldn't this be in Utils?

Assets/Scripts/Crafting/CraftingRecipeUI.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// <param name="recipe">The recipe of this crafting recipe UI</param>
+        /// <param name="craftingUI">The crafting UI that should hold the correct <see cref="CraftingUI.ItemHandler"/></param>
+        public void InitRecipe(CraftingRecipe recipe, CraftingUI craftingUI)
+        {
+            _craftingUI = craftingUI;
+            _recipe = recipe;
+
+            txtTitle.SetText(_recipe.CreatedItemName);
+            foreach (CraftingRecipe.ResourceData resourceData in _recipe.NeededItems)
+                Instantiate(craftingRecipeResourcePrefab, recipeResourceList.transform)
```
SamuelllG, line 50: this is for displaying the things you need for crafting, right?

Assets/Scripts/Crafting/CraftingRecipeUI.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// The crafting UI reference that holds the item handler
+        /// </summary>
+        private CraftingUI _craftingUI;
+
+        /// <summary>
+        /// The recipe of this CraftingRecipeUI element
+        /// </summary>
+        private CraftingRecipe _recipe;
+
+        private void Start() => OnItemUpdate();
```
SamuelllG, line 36: this is called every time you open the crafting UI ? so it doesn't have to be updated? (i.e. when you gather a missing resource for the recipe, and the recipe should be craftable)

Assets/Scripts/Entity/Player/PlayerController.cs
```diff
@@ -17,38 +19,37 @@ public class PlayerController : Movable
...
         private LayerMask clickableLayers;
-        
+
         [Tooltip("Defines where the player can move")] [SerializeField]
         private LayerMask groundLayer;
 
 
-        [Tooltip("An effect that will be displayed whenever the player clicks to move")]
-        [SerializeField]
+        [Tooltip("An effect that will be displayed whenever the player clicks to move")] [SerializeField]
```
SamuelllG, line 27: haha, your work will be destroyed as soon as someone uses auto format :[

Assets/Scripts/Crafting/CraftingRecipeUI.cs
SamuelllG: ah didn't see ItemsUpdated l. 54 in InitRecipe :)

SamuelllG (dismissed):
wow, I'm **delighted** haha 👍 I literally just found a missing "w" in a comment
Added a few (dumb) questions, and you might put the struct to Utils, idk

Assets/Scripts/Crafting/CraftingRecipe.cs
DavidPgl: I could put it there but it's pretty specific and will only be used with the CraftingRecipes. It's more like a helper class for the recipes and keeping them together is a bit more organized I think. But I can move it as well:)

Assets/Scripts/Crafting/CraftingRecipeUI.cs
DavidPgl: Exactly, it spawns in each resource

Assets/Scripts/Crafting/CraftingRecipeUI.cs
DavidPgl: This is only for triggering the update event on start. Makes sure it turns red from the beginning if you can't craft it and will be helpful later on when you load the game with items in your inventory

Assets/Scripts/Entity/Player/PlayerController.cs
DavidPgl: This was auto-formatting, so no worries:D

Assets/Scripts/UI/InventoryUI.cs
```diff
@@ -1,10 +1,13 @@
...
 using System.Collections.Generic;
+using System.Linq;
+using Interfaces;
+using Inventory;
 using UnityEngine;
 
 namespace UI
 {
-    public class InventoryUI : MonoBehaviour
+    public class InventoryUI : MonoBehaviour, IItemHandler
```
Kaes3kuch3n: I think `IItemHandler` should be implemented by the `Inventory` class itself, and not by `InventoryUI` 🤔 Afaik the inventory class also already defines item update events, so you could either build on that foundation or rewrite it so that it fits the interface :)

Assets/Scripts/AbstractClasses/IItemHandler.cs
Kaes3kuch3n: In my opinion this method should also define a default value for the amount of items, for consistency reasons :)

Kaes3kuch3n (dismissed):
Nice 👌

Kaes3kuch3n (changes requested):
Okay, 2nd try: See comments above 😅

Kaes3kuch3n (approved):
👌🏻

inkompotato (approved):
looking good, took some time until I found out about the little button to add items.

cuongtm95:
Found a bug. When we craft stuffs, for example the pick axe, which requires 3x stone and 2x wood. After crafting, the resources were removed from UI but not be fully removed from the Inventory Controller, instead only 1 of them did. So next time we want to craft the same item, we only need 1 more stone and 1 more wood, although it doesn't show anything on the UI, because the value still being stored in the inventory list.
crafting-bug

PR #66, opened 2019-12-14
NPC spawning

SamuelllG:
- new prefab NPCspawner
- no NPCs spawned if in camera frustum
- if in camera frustum but behind wall, NPCs are still NOT spawned, because culling is not working
- for this implemented RenderChecker which uses an empty test renderer.. if culling works, just add a test renderer to the enemy prefab, enable the RenderChecker on the enemy prefab, and delete some lines in TrySpawn() in NPCSpawner script
- reused wanderAi methods and made NavMeshMapper
- added tag for NPCs
- made WanderArea reusable
- implemented probability utility class


Assets/Scripts/Entity/Enemy/EnemyController.cs
```diff
@@ -12,15 +12,15 @@ public class EnemyController : Movable
...
+        private Area wanderArea;
 
         [Tooltip("The frequency at which the NPC walks to a new point in the wander area.")]
         [SerializeField]
         private float wanderTimer;
 
         [Tooltip("Whether the area should be fixed on enable.")]
         [SerializeField]
-        public bool freezeArea;
+        private bool freezeArea;
```
DavidPgl, line 23:
```suggestion
        private bool _freezeArea;
```

Assets/Scripts/Entity/Enemy/RenderChecker.cs
```diff
@@ -0,0 +1,48 @@
...
+    /// Otherwise activate real mesh renderer.
+    /// </summary>
+    private void OnWillRenderObject()
+    {
+        if (_isNew)
+        {
+            _isNew = false;
+            if (_spawnRenderTimer < 0.1)
+            {
+                Destroy(transform.gameObject);
```
DavidPgl:
```suggestion
                Destroy(this.gameObject);
```
Should work as well :D

Assets/Scripts/Spawner/NPCSpawner.cs
```diff
@@ -0,0 +1,100 @@
...
+        private int maximumConcurrentEnemies;
+
+        [Tooltip("The probability to spawn an NPC for each frame")]
+        [SerializeField]
+        [Range(0, 0.1f)]
+        private float spawnProbability;
+
+        [Tooltip("The NPC to spawn")]
+        [SerializeField]
+        private GameObject toSpawn;
```
DavidPgl: You can use the exact type here as well. E.g. EnemyController. With this you can access for example the mesh renderer of it and don't need to use the _**forbidden**_ GetComponentChild :P

Assets/Scripts/Spawner/NPCSpawner.cs
```diff
@@ -0,0 +1,100 @@
...
+        /// <summary>
+        /// Try spawning an NPC. The NPC might not consist in the scene 
+        /// (but will be destroyed before it is made visible if it's in the player's sight).
+        /// </summary>
+        private void TrySpawn()
+        {
+            Vector3 spawnLocation = _navMeshMapper.GetMappedRandomPoint(spawnArea, spawnLayer.value);
+            // setting a random rotation doesn't seem to work with a NavMeshAgent, but since NPCs start walking in a random direction immediately it doesn't really matter
+            GameObject spawned = Instantiate(toSpawn, spawnLocation, Quaternion.identity, transform);
+            MeshRenderer renderer = spawned.GetComponentInChildren<MeshRenderer>();
```
DavidPgl: As mentioned before. Use exact type

Assets/Scripts/Utils/Probability.cs
```diff
@@ -0,0 +1,45 @@
...
+    public class Probability
+    {
+        private Random _random;
+
+        public Probability() => _random = new Random();
+
+        /// <summary>
+        /// Returns true with a certain probability.
+        /// </summary>
+        /// <param name="probability">The probability to return true</param>
```
DavidPgl: Add here that the value needs to be in-between 0 and 1

DavidPgl (changes requested):

Assets/Scripts/Entity/Enemy/EnemyController.cs
SamuelllG: this is a SerializeField

Assets/Scripts/Entity/Enemy/RenderChecker.cs
SamuelllG: Where is the difference then? :D 
also == Destroy(gameObject)

Assets/Scripts/Spawner/NPCSpawner.cs
SamuelllG: what do you mean? In toSpawn I put the prefab that should be spawned.. how should I put EnemyController here and how should this help with GetComponent in child?

SamuelllG: I know GetComponentInChildren is ugly but I couldn't think of a better solution 😔 The alternative is to do it with the RenderChecker, but since this is a lot of stuff, I preferred the other way

Assets/Scripts/Spawner/NPCSpawner.cs
SamuelllG: But I literally instantiate a _gameObject_ , don't I?

Assets/Scripts/Utils/ClassExtender.cs
```diff
@@ -59,5 +59,17 @@ public static class ClassExtender
...
             return null;
         }
+
+        /// <summary>
+        /// Checks if an object is visible from camera's perspective.
+        /// </summary>
+        /// <param name="renderer">The renderer of the gameobject to be checked</param>
+        /// <param name="camera">The camera to use</param>
+        /// <returns></returns>
+        public static bool InFrustum(this Renderer renderer, Camera camera)
```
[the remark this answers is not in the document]
SamuelllG: 😂

Assets/Scripts/Spawner/NPCSpawner.cs
DavidPgl: Instantiate is a generic method:
`public T Instantiate<T>()`
which means it returns the type you give it. You **should** be able to define it as EnemyController and drag the prefab in. Then, when instantiating, you automatically get the type back and you could call a reference from it e.g. `enemy.Renderer`

Assets/Scripts/Entity/Enemy/RenderChecker.cs
```diff
@@ -0,0 +1,48 @@
+﻿using UnityEngine;
+
+/// <summary>
+/// Currently objects are rendered also behind walls, so the desired effect, 
+/// that NPCs do spawn as long as the player can't see them (so in camera frustum but behind walls as well), is not achieved.
+/// So this is not used, currently.
+/// </summary>
+public class RenderChecker : MonoBehaviour
```
DavidPgl: Namespace:)

Assets/Scripts/Entity/Enemy/RenderChecker.cs
```diff
@@ -0,0 +1,48 @@
...
+{
+    [Tooltip("The renderer that renders the real mesh")]
+    [SerializeField]
+    private MeshRenderer _meshRenderer;
```
DavidPgl:
```suggestion
    private MeshRenderer meshRenderer;
```

Assets/Scripts/Utils/Probability.cs
```diff
@@ -0,0 +1,45 @@
...
+            if (percentage >= 1)
+            {
+                return temp < percentage;
+            }
+            else // < 1% -----> new min: 0.01%
+            {
+                if (temp == 0) // hit the one percent
+                {
+                    // 1 old percent = 100 new percent
+                    return GetProbability(percentage);
```
DavidPgl, line 39: Won't this throw an exception every time? It takes the percentage value (which is between 0 and 100) and calls the same method again. Also, I'm not sure if you should call this recursive like this. Maybe you should calculate the value you multiply the probability with, with the probability you get? E.g. 0.01 - 100 / 0.001 - 1000? You could also use floats for the random as well I think

DavidPgl (changes requested):
Last bit, mostly questions

Assets/Scripts/Spawner/NPCSpawner.cs
SamuelllG: Very good idea!

Assets/Scripts/Entity/Enemy/RenderChecker.cs
SamuelllG: damn, the renderchecker xD

Assets/Scripts/Utils/Probability.cs
SamuelllG: It doesn't throw an exception because it's in the else that is only entered if percentage < 1 (e.g. 0.8).

> calculate the value you multiply the probability with, with the probability you get

I'm not exactly sure how you mean it, because the method simply doesn't return `true` with the correct probability if you multiply the passed probability with a value depending on the dimension of the probability. 😬 I like the recursion because it's not too difficult to read (and it works 😁 ). 
If you have a better solution, you can implement it of course.

DavidPgl: it's fine like that:)

DavidPgl (approved):

thuemmels (approved):

PR #71, opened 2019-12-18
Sanity fog of sanity

tobihoffmann:


DavidPgl:
image

So this is in the GameScene. Can you remove it and add the references to the Correct one?
Also, why is it on the Player? Could you move it to the FogOfSanityPrefab?

Assets/Scripts/Effects/FogOfSanity.cs
```diff
@@ -0,0 +1,100 @@
...
+    private float _currentRadius;
+    private bool _isGrowing;
+   
+    // shader properties
+    private static readonly int PlayerPosition = Shader.PropertyToID("_PlayerPosition");
+    private static readonly int FogRadius = Shader.PropertyToID("_FogRadius");
+
+    private void Awake()
+    {
+        _baseRadius = 500;
```
DavidPgl: Do we need to change this value depending on the screensize?

tobihoffmann:
> image
> 
> So this is in the GameScene. Can you remove it and add the references to the Correct one?
> Also, why is it on the Player? Could you move it to the FogOfSanityPrefab?

Are you seeing this component twice on Player? This might be an editor bug then b/c i only see it once. I put it on Player since the center of it is attached to Player position. But i could move this to FogOfSanity if you want

Assets/Scripts/Effects/FogOfSanity.cs
tobihoffmann: this value is relative to clip space coordinates, so i'd say no. But i'd have to jump deeper into shader logic to say for certain. However it's scaling mechanic is pretty basic at the moment and will need some fine tuning, once we completed all the visual stuff that is connected to the sanity system anyway.

Assets/Scripts/Effects/FogOfSanity.cs
```diff
@@ -0,0 +1,100 @@
+﻿using Entity.Player;
+using UnityEngine;
+
+
+public class FogOfSanity : MonoBehaviour
+{
+    [SerializeField][Tooltip("Reference to the fog generating mesh")]
+    private Renderer fogOfSanityMesh;
+    
+    [SerializeField][Tooltip("Reference to the fog generating mesh")]
```
Kaes3kuch3n:
```suggestion
    [SerializeField][Tooltip("Main camera in the scene")]
```

Kaes3kuch3n (commented):
Nur ne Kleinigkeit :) Ansonsten echt nice 👌

DavidPgl:
> > image
> > So this is in the GameScene. Can you remove it and add the references to the Correct one?
> > Also, why is it on the Player? Could you move it to the FogOfSanityPrefab?
> 
> Are you seeing this component twice on Player? This might be an editor bug then b/c i only see it once. I put it on Player since the center of it is attached to Player position. But i could move this to FogOfSanity if you want

I would like it on FogOfSanity:)

Kaes3kuch3n (approved):

DavidPgl (approved):

PR #73, opened 2019-12-24
ItemsStacking

DavidPgl:


thuemmels (changes requested):
Code looks good, but there were some bugs during testing that I still don't fully understand.
First of all, item swapping and stacking seem to work just fine when using only items that were added to the inventory via the editor. But if you move some of these items around and then craft something, things get weird. The buttons can move out of the inventory space or arrange in a new order.
And when you move crafted items around (e.g. pickaxes), it seems like the buttons are overlapping. Once I had multiple pickaxes in one grid cell and thought they disappeared, but I was able to seperate them again. It's also worth noting that this behaviour only happens one time, and after that the crafted item behaves normally again.

Nice to have:
Currently, if you have multiple pickaxes in your inventory, the first one in the list is the one going to be equipped, no matter what button to press on. Would be nice if we could change that to "I want to equip the specific item that I clicked on."

DavidPgl:
Damn that was fast. Wanted to test it tomorrow😅
Will fix it and test some more:)

PR #76, opened 2019-12-28
Bugfix: NPC spawner

Kaes3kuch3n:
Removed layer mask field from spawner, since the nav mesh mapping function doesn't require a layer mask, but an area mask for the nav mesh instead. Sadly, Unity doesn't provide a class for this, so we would need to write our own inspector for this to work properly. But since I don't think that we need this anytime soon, I've just removed it for now.

### DIFF
--- a/Assets/Scripts/Entity/Player/PlayerController.cs
+++ b/Assets/Scripts/Entity/Player/PlayerController.cs
@@ -144,38 +144,36 @@ namespace Entity.Player
         {
             if (EventSystem.current.IsPointerOverGameObject())
                 return;
-            else
+            
+            Ray clickRay = _camera.ScreenPointToRay(Mouse.current.position.ReadValue());
+
+            // only change target / move, if not performing a hit
+            if (Physics.Raycast(clickRay, out RaycastHit hit, 10000, clickableLayers) &&
+                _attackLogic.Status == AttackLogic.AttackStatus.None)
             {
-                Ray clickRay = _camera.ScreenPointToRay(Mouse.current.position.ReadValue());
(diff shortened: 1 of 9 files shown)


Kaes3kuch3n:
Did some more cleanup and refactoring related to the NPC scripts :)

Assets/Scripts/Entity/Player/PlayerController.cs
```diff
@@ -144,38 +144,36 @@ private void OnRightClick(InputAction.CallbackContext obj)
...
+            Ray clickRay = _camera.ScreenPointToRay(Mouse.current.position.ReadValue());
+
+            // only change target / move, if not performing a hit
+            if (Physics.Raycast(clickRay, out RaycastHit hit, 10000, clickableLayers) &&
+                _attackLogic.Status == AttackLogic.AttackStatus.None)
             {
-                Ray clickRay = _camera.ScreenPointToRay(Mouse.current.position.ReadValue());
+                GameObject objectHit = hit.collider.gameObject;
+
+                if (objectHit.GetComponent<Damageable>() is Damageable damageable)
```
DavidPgl: TryGetComponent?

DavidPgl (changes requested):
Looks nice, just one question.

Assets/Scripts/Entity/Player/PlayerController.cs
Kaes3kuch3n: Both options work fine, but since we use TryGetComponent in other places, too, I've changed it for more consistency

DavidPgl (approved):

SamuelllG (approved):

PR #77, opened 2019-12-29
Sanity fog of sanity

tobihoffmann:


Kaes3kuch3n (dismissed):

Assets/Scripts/Effects/FogOfSanity.cs
```diff
@@ -94,7 +89,7 @@ private void UpdateFogRadius(int sanityLevel)
             }
             else
             {
-                _baseRadius = sanityLevel * 2;
+                _baseRadius = sanityLevel * 1.5f;
```
DavidPgl: Could this be a serialized field, so we can tweak it later on?

DavidPgl (changes requested):
Just one really small thing:)

Assets/Scripts/Effects/FogOfSanity.cs
tobihoffmann: ✓

DavidPgl (approved):

SamuelllG (approved):

PR #81, opened 2020-01-02
Sanity mobs

SamuelllG:
Spawned sanity mobs are visible for all players for now.

I tried to put the sanity spawner into the player prefab, but it's not possible because the NavMeshAgent can only be rotated by setting transform.rotation directly which then causes the spawner (and all spawned mobs) to rotate with the player whenever he's trying to face a target (i.e. rotating while standing).

SanitySpawner: Spawns mobs around the player depending on the sanity level. There is a sanity mob prefab. Sanity mobs should not be killable (or we have to change some logic). Thus they despawn after a certain time depending on the player's sanity. Sanity mobs follow the player and despawn automatically when they're out of sight.

If you want to try it out, just set a lower sanity using the remote and wait a while :)


SamuelllG:
I pushed changes of the gamescene in one commit. Not sure about what we said though... but I guess you can easily revert a commit with gamescene changes only, sooo ...

Assets/Scripts/Spawner/SanitySpawnController.cs
```diff
@@ -0,0 +1,119 @@
...
+using Utils;
+using System.Linq;
+using System.Collections.Generic;
+using Boo.Lang;
+
+namespace Spawner
+{
+    public class SanitySpawnController : NPCSpawner
+    {
+        public delegate void PlayerStateChanged(int newValue);
```
DavidPgl: Do you use this?

Assets/Scripts/Spawner/SanitySpawnController.cs
```diff
@@ -0,0 +1,119 @@
...
+
+        [SerializeField]
+        private AnimationCurve probabilityScaleCurve;
+
+        [Tooltip("Spawn probability is scaled with a factor 1-x for every active/existing sanity NPC.")]
+        [SerializeField]
+        private float activeSpawnPenalty;
+
+
+        private Dictionary<Renderer, GameObject> _spawned;
```
DavidPgl:
```suggestion
        private Dictionary<Renderer, GameObject> _spawned = new Dictionary<Renderer, GameObject>();
```
No need to do it in Awake. Also, if you never overwrite it with a new dictionary, you can add the keyword `readonly` as well.

Assets/Scripts/Spawner/SanitySpawnController.cs
```diff
@@ -0,0 +1,119 @@
...
+            renderer.enabled = true;
+            _spawned.Add(renderer, spawned.gameObject);
+            /*if (!renderer.InFrustum(mainCamera))
+                renderer.enabled = true;
+            else
+                Destroy(spawned.gameObject);*/
+            spawned.WanderArea.center = spawnArea.center;
+            CalculateSpawnProbability(_sanity);
+            //base.TrySpawn();
+            //spawned.WanderArea.center = spawnArea.center;
```
DavidPgl: All this commented out code still needed?

DavidPgl (changes requested):

SamuelllG:
SORRY, I forgot to do a lookover... will add descriptions, etc. !

DavidPgl (approved):

thuemmels (approved):

PR #82, opened 2020-01-03
Npc lifecycle

SamuelllG:
Health, Dying and regeneration implemented

### DIFF
--- a/Assets/Scripts/Entity/Enemy/EnemyState.cs
+++ b/Assets/Scripts/Entity/Enemy/EnemyState.cs
@@ -2,25 +2,61 @@
 using System.Collections;
 using System.Collections.Generic;
 using UnityEngine;
+using Utils;
 
 namespace Entity.Enemy {
     
     public class EnemyState : Damageable
     {
-        [Tooltip("100 - full health, 0 - dead")]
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Entity/Enemy/EnemyState.cs
```diff
@@ -2,17 +2,81 @@
...
+            health = updatedValue;
+        }
+
+        private void Regenerate()
+        {
+            if (_probability.GetProbability(regenerationProbability))
+            {
+                System.Random random = new System.Random();
+                float x = (float) random.NextDouble();
+                Debug.Log("x: " + x);
```
DavidPgl: If not needed, remove debugs

DavidPgl (changes requested):
No summaries?:(

SamuelllG:
SORRY, I forgot to do a lookover... will add descriptions, etc. !

DavidPgl (approved):

thuemmels (approved):
I think using probabilities is a bit weird when it comes to regeneration, but it works, so why not :D

SamuelllG:
I used probabilities so that the enemy doesn't regenerate consistently

PR #83, opened 2020-01-03
Audio manager

tobihoffmann:
To be discussed:
atm all audio sources (components) will be created through start() of audio manager since singleton uses awake. What would be a better solution?
Maybe let AudioManager have its own singleton method?
Or creating the audio sources pre runtime? (I have never done this)

Also all fields in Sound are public, do we want to make them properties?


Assets/Scripts/Managers/AudioManager.cs
```diff
@@ -0,0 +1,107 @@
...
+     */
+    public class AudioManager : Singleton<AudioManager>
+    {
+        [SerializeField][Tooltip("Container for all sounds")]
+        private List<Sound> sounds;
+
+        [SerializeField]
+        private GameObject audioSources;
+
+        private void Start()
```
DavidPgl:
```suggestion
        protected override void Awake()
        {
            base.Awake():
        }
```
You can override the awake function of the singleton. Make sure to call `base.Awake()`

DavidPgl (changes requested):

Assets/Scripts/Managers/AudioManager.cs
tobihoffmann: thanks, will try!

DavidPgl (approved):

inkompotato (approved):
nice, final game sounds coming soon?

PR #93, opened 2020-01-08
Day events

tobihoffmann:
Added events for sunrise and sunset mainly to trigger atmospheric audio.
We could also use them to fire status updates like "Another day survived ..." or else.

### DIFF
--- a/Assets/Scripts/GameTime/Clock.cs
+++ b/Assets/Scripts/GameTime/Clock.cs
@@ -51,6 +51,7 @@ namespace GameTime
         {
             UpdateTimeScale();
             UpdateTime();
+            InvokeSunRiseAndSunSet();
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


Assets/Scripts/GameTime/Clock.cs
```diff
@@ -6,6 +6,10 @@ namespace GameTime
 {
     public class Clock : MonoBehaviour
     {
+        public delegate void SunRise();
```
Kaes3kuch3n: I'd suggest just creating one delegate called something like `SunStatus` or `DayTimeEvent` and using this for both events

Assets/Scripts/GameTime/Clock.cs
```diff
@@ -87,5 +116,13 @@ private void NormalizeTimeCurve()
 
             _timeCurveNormalization = curveTotal / numberOfSteps;
         }
+        
+        /// <summary>
+        /// Checks if it is day or night and sets "_isNight" accordingly
+        /// </summary>
+        private void SetDayNightTriggers()
+        {
+            _isNight = (TimeOfDay >= 0.25 && TimeOfDay <= 0.75) ? _isNight = false : _isNight = true;
```
Kaes3kuch3n:
```suggestion
            _isNight = !(TimeOfDay >= 0.25 && TimeOfDay <= 0.75);
```

Kaes3kuch3n (changes requested):

Assets/Scripts/GameTime/Clock.cs
tobihoffmann: what is this dark magic?

DavidPgl (approved):

Kaes3kuch3n (approved):

PR #94, opened 2020-01-09
Buildings v2

DavidPgl:


Assets/Scripts/Buildings/Building.cs
```diff
@@ -26,44 +29,56 @@ public class Building : MonoBehaviour
...
 
-            if (PlayerInReach && !_playerWasInReach)
-            {
-                _playerWasInReach = true;
-                CraftingManager.Instance.CurrentCraftingStation = craftingStation;
-            }
+        /// <summary>
+        /// Can be overriden to define an interaction action.
+        /// </summary>
+        public virtual void Interact()
```
Kaes3kuch3n, line 73: I think this can be defined as abstract, if it doesn't contain any code :)

Assets/Scripts/Buildings/Building.cs
```diff
@@ -26,44 +29,56 @@ public class Building : MonoBehaviour
...
-            // Walked away from this crafting station
-            if (!PlayerInReach && _playerWasInReach)
-            {
-                _playerWasInReach = false;
-                CraftingManager.Instance.CurrentCraftingStation = CraftingManager.CraftingStation.None;
-            }
+        /// <summary>
+        /// Can be overriden to define an action on building this building.
+        /// </summary>
+        protected virtual void OnBuild()
```
Kaes3kuch3n, line 80: I think this can be defined as abstract, if it doesn't contain any code :)

Assets/Scripts/Buildings/BuildingScripts/Tent.cs
```diff
@@ -0,0 +1,34 @@
...
+        /// Sets the spawn point after the player finished building it.
+        /// </summary>
+        protected override void OnBuild() => PlayerManager.Instance.SetSpawnPoint(spawnPoint.transform);
+
+        /// <summary>
+        /// Changes the time of the day to the specified one using the <see cref="DayNightManager"/>.
+        /// </summary>
+        public override void Interact()
+        {
+            if (!PlayerInReach) return;
```
Kaes3kuch3n, line 30: I guess this could be moved to the base class as well. Should be the same for every building, right?

Assets/Scripts/GameTime/Clock.cs
```diff
@@ -50,6 +52,22 @@ private void Update()
             InvokeSunRiseAndSunSet();
         }
 
+        /// <summary>
+        /// Sets the time of the day instantly. Allows values between 0 and 2 (Everything over 1 being the next day).
+        /// </summary>
+        /// <param name="newTimeOfDay">The new time of the day (0 to 2)</param>
+        public void SetTimeOfDay(float newTimeOfDay)
+        {
+            if (newTimeOfDay >= 1)
```
Kaes3kuch3n, line 61: Move to animation please :)

Kaes3kuch3n (changes requested):

PR #101, opened 2020-01-11
Buildings v2

DavidPgl:


Assets/Scripts/Buildings/Building.cs
```diff
@@ -4,66 +4,70 @@
...
+        /// Whether the building is build or not. Invokes <see cref="OnBuild"/>
         /// </summary>
-        public bool IsBuild { get; set; }
+        public bool IsBuild
+        {
+            get => _isBuild;
+            set
+            {
+                _isBuild = value;
+                OnBuild();
```
Kaes3kuch3n: Could you add a check to see if the property is actually set to `true` before executing `OnBuild()`?

Assets/Scripts/Buildings/Campsite.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// <summary>
+        /// Whether this campsite is already unlocked.
+        /// </summary>
+        public bool IsUnlocked
+        {
+            get => isUnlocked;
+            set
+            {
+                isUnlocked = value;
+                OnUnlock?.Invoke();
```
Kaes3kuch3n: Please also add a check here (see above)

Assets/Scripts/Managers/PlayerManager.cs
```diff
@@ -5,7 +5,11 @@ namespace Managers
     public class PlayerManager : Singleton<PlayerManager>
     {
         [SerializeField] private GameObject player;
-        
+
         public GameObject GetPlayer() => player;
+
+        public bool PlayerInReach(GameObject otherObject, float distanceToPlayer) =>
```
Kaes3kuch3n: Maybe rename this to something like `reachDistance` or `maxRange` to make it clearer what's actually meant? And/Or add documentation :P

Kaes3kuch3n (dismissed):

Assets/Scripts/Buildings/Building.cs
DavidPgl: Hahaha, that's how it should work 😂 Will do

SamuelllG (approved):
👌

inkompotato (approved):

PR #102, opened 2020-01-11
Animation variants

Kaes3kuch3n:


SamuelllG (approved):
why are all the commits of the previous animation PR in here as well?

DavidPgl (approved):

Kaes3kuch3n:
> why are all the commits of the previous animation PR in here as well?

Because I created this branch from the player model branch before it was merged to master, and GitHub isn‘t smart enough to filter out commits that are already in master @SamuelllG 😅

PR #103, opened 2020-01-11
Items v2

DavidPgl:


SamuelllG (dismissed):
don't really have much to say (not that that's a bad thing) 😬 nice work

Kaes3kuch3n (dismissed):

Assets/Scripts/Managers/InventoryManager.cs
```diff
@@ -19,33 +20,56 @@ public class InventoryManager : Singleton<InventoryManager>
...
         /// </summary>
         /// <param name="item">The item to be added</param>
         /// <param name="amount">The amount</param>
-        public void AddItem(Item item, int amount = 1) => playerInventory.AddItem(item, amount);
+        public bool AddItem(Item item, int amount = 1)
+        {
+            for (int i = 0; i < amount; i++)
+            {
+                if (!HasEmptySlots && !NotFullItemStacks.Contains(item)) return false;
+                playerInventory.AddItem(item);
```
Kaes3kuch3n: May add some items and still return false if not all items could be added. Is this intended/do you have that in mind when calling the method? Might wanna add some info to the doc here :)

DavidPgl: Oh damn, you're right

Kaes3kuch3n (dismissed):

Kaes3kuch3n (approved):

SamuelllG (approved):

PR #108, opened 2020-01-14
Item hover tooltip

DavidPgl:


DavidPgl:
So, @cuongtm95 and me accidentally worked on the same stuff. We can use his version of the UI-Hover as well.
Haven't done Crafting but, using this as a base, it should rather be a question of design than of logic or implementation

cuongtm95:
Just checked out your work. Great job! Mine has the same logic but yours is for sure more detailed and has a perfect format (not to mention some other bonus stuff you added there for the inventory). Gonna delete mine and use yours! 👍 💯 

For crafting window, I find it better looking with just the icon of the item/recipe and display the tooltip on hover that shows the needed resources and some description. Or even better if we can create a new window that appears when u click on a recipe. The new window could show the item name, description, resources counters and a button "Craft" at the end. No doubt that will need more works but yeah just some thoughts.

Assets/Scripts/Utils/ElementInteraction/Draggable.cs
```diff
@@ -11,41 +12,64 @@ namespace Utils.ElementInteraction
...
+        [Tooltip("Used for showing this on top of other UI-Elements")] [SerializeField]
+        private Canvas canvas;
+
+        [Tooltip("Used for finding UI elements with rays")] [SerializeField]
+        private GraphicRaycaster graphicRaycaster;
+
         [Tooltip("The highest parent element that will be moved")] [SerializeField]
         private Transform parent;
 
+        public GraphicRaycaster GraphicRaycaster { get; private set; }
```
Kaes3kuch3n: You are currently using two different graphic raycasters: One which is stored in this property and fetched using GetComponentsInParent on Awake, and one which is stored in a serialised field. I guess that's not intended

Kaes3kuch3n (changes requested):
Just one comment :)

Assets/Scripts/Utils/ElementInteraction/Draggable.cs
```diff
@@ -11,41 +12,64 @@ namespace Utils.ElementInteraction
...
+            }
+        }
+
         /// <summary>
         /// Gets the GraphicRayCaster from the canvas
         /// </summary>
         private void Awake()
         {
-            // Get the closest GraphicRaycaster (in hierarchy)
-            GraphicRaycaster = GetComponentInParent<GraphicRaycaster>();
```
SamuelllG, line 40: 😵 that's FORBIDDEN 😁

Assets/Scripts/Utils/ElementInteraction/Draggable.cs
SamuelllG: Hmm why do you use GetComponentsInParent anyway if you have a serialized field?

SamuelllG (changes requested):
Looks good, except for the FORBIDDEN GetComponentsInParent 😝

Assets/Scripts/Utils/ElementInteraction/Draggable.cs
DavidPgl: I DONT CARE 🚽

Assets/Scripts/Utils/ElementInteraction/Draggable.cs
DavidPgl: Renamed it

Kaes3kuch3n (dismissed):

Kaes3kuch3n (approved):

SamuelllG (approved):

PR #118, opened 2020-01-15
Fixed automatic entity rotation

Kaes3kuch3n:
Rotation should be working now, and it also looks a little better than before in my opinion :)


SamuelllG (approved):
Didn't test it though

DavidPgl (approved):

PR #120, opened 2020-01-15
Added entity sounds

Kaes3kuch3n:


Assets/Scripts/Entity/EntitySounds.cs
```diff
@@ -0,0 +1,22 @@
+using UnityEngine;
+
+namespace Entity
```
tobihoffmann: Could put this into GameAudio

SamuelllG (dismissed):

Assets/Scripts/Entity/EntitySounds.cs
Kaes3kuch3n: Ah, didn't notice we have a namespace specifically for audio 👍

tobihoffmann (approved):

SamuelllG (approved):

PR #132, opened 2020-01-16
Fire particles and damage

Kaes3kuch3n:


Assets/Scripts/Buildings/Campfire.cs
```diff
@@ -0,0 +1,35 @@
...
+        private int damage;
+
+        [Tooltip("The interval in which to apply the damage to the player")] [SerializeField]
+        private float damageInterval;
+
+        private float _timer;
+
+        private void OnTriggerStay(Collider other)
+        {
+            if (!other.tag.Equals(Consts.Tags.PLAYER)) return;
```
DavidPgl:
```suggestion
            if (!other.CompareTag(Consts.Tags.PLAYER)) return;
```
If you do it with OnTriggerStay, then use this. It's a bit faster:P

Assets/Scripts/Buildings/Campfire.cs
```diff
@@ -0,0 +1,35 @@
...
+            
+            _timer += Time.deltaTime;
+            if (_timer < damageInterval) return;
+            
+            other.GetComponent<PlayerState>().Hit(damage);
+            _timer = 0;
+        }
+
+        private void OnTriggerExit(Collider other)
```
DavidPgl, line 33: If you use OnTriggerExit, couldn't you use OnTriggerEnter as well and save the PlayerState as a variable. And while this variable is not null you apply damage (with the damageInterval) in Update. Then you would save a lot of Tag comparing and GetComponent calls

DavidPgl (changes requested):

inkompotato (dismissed):

Assets/Scripts/Buildings/Campfire.cs
Kaes3kuch3n: Added reference in awake, since we aren't gonna support multiplayer anytime soon anyway :)

inkompotato (approved):

DavidPgl (approved):

PR #137, opened 2020-01-17
Interact Tooltip Improved 

cuongtm95:
### What's changed:
- Renamed the script to **InteractTooltip**
- The tooltip will now auto resized relative to camera distance (the tooltip getting bigger when we approaching from afar in first person pov and vice versa).
- Added new function: **SetCursor**. The cursor will be set to a custom texture sprite (must be cursor type in unity import setting) when hovering over the object.
- Be able to set the position and size of the tooltip icon in the inspector.
- Removed some unused img files.

### Thoughts:

Now we have 3 ways to display the InteractTooltip:
- Text Tooltip with an epic font (like Don't Starve did)
- Custom icon (show only the icon/image/texture, still have the main cursor) 
- Custom cursor

Ofc we can combine any of the 3 together, just need to tweak some values in the inspector.

### Issue:
- Not yet found a way to automatically calculate the offset position for the icon to correctly be displayed beside the text (when using both function 1 and 2)


Assets/Scripts/UI/InteractTooltip.cs
```diff
@@ -161,16 +165,15 @@ private void OnGUI()
             // Display of text/tooltip that follows the mouse
             if (!hideTooltip && followingMouseCursor && _over)
             {
-                if (textResized)
+                Debug.Log(_tooltipStyle.lineHeight);
```
DavidPgl: Remove if not needed

DavidPgl (changes requested):

Assets/Scripts/UI/InteractTooltip.cs
cuongtm95: oh yeah was forgot to remove this line. used for testing some stuff

SamuelllG (approved):

DavidPgl (approved):

PR #144, opened 2020-01-18
fixes

SamuelllG:
fixed abort attack
fixed target approaching bugs
fixed stopping distance = 0
improved attack logic slightly

tried to implement exact approaching, but it turns out it doesn't really matter if I check the collider extent of the target in the walking direction and subtract them from my move position because with the navmeshagent it doesn't make a real difference

I adjusted the too big navmeshagent obstacle avoidance radius' though which fixed the bug


DavidPgl (dismissed):

cuongtm95 (dismissed):

SamuelllG:
Had some problem merging, please re-review

DavidPgl (approved):

cuongtm95 (approved):

PR #147, opened 2020-01-19
SAVE our games!!

inkompotato:
adds the ability to save the current game state via remote console...
use the following remote console commands for testing:
`save/test` will create a file called `TEST.json` at the Application Persistent Data Path, which is in `appdata/loacallow/defaultcompany/insanesurvival` on windows. 

Load a file with `load/[filename]`

The save-games are not intercompatible between different builds of the game, including identical editor- and build- versions.

The following things are saved:

- World State (time / daynumber)
- Player Inventory
- Campsites & Buildings
- Harvestables
- Player State

A lot of files were touched, as I needed to modify them to be serializable and support setting and getting the needed data. When reviewing, please test the saving functionality as described above and report any issues here...


Assets/Scripts/Entity/Player/PlayerState.cs
```diff
@@ -41,10 +41,10 @@ public class PlayerState : Damageable
         private void Awake()
         {
             // send event on initial values
-            OnPlayerHealthUpdate?.Invoke(health);
+            /*OnPlayerHealthUpdate?.Invoke(health);
             OnPlayerSaturationUpdate?.Invoke(saturation);
             OnPlayerHydrationUpdate?.Invoke(hydration);
-            OnPlayerSanityUpdate?.Invoke(sanity);
+            OnPlayerSanityUpdate?.Invoke(sanity);*/
```
DavidPgl: If not needed delete I guess?

Assets/Scripts/Interactables/Harvestable.cs
```diff
@@ -71,7 +87,7 @@ public override void Interact()
             else
             {
                 // TODO: Show needed item.
-                Debug.Log("Missing ability");
+                Debug.Log("Missing ability " +neededAbility);
```
DavidPgl: Remove if not needed

Assets/Scripts/Interactables/Harvestable.cs
```diff
@@ -128,6 +144,8 @@ private void AddItems()
         /// </summary>
         private IEnumerator Respawn()
         {
+            Debug.Log("respawn started");
```
DavidPgl: Remove debugs in general

Assets/Scripts/Managers/SaveManager.cs
```diff
@@ -0,0 +1,173 @@
...
+{
+    public class SaveManager : Singleton<SaveManager>
+    {
+        /// <summary>
+        /// saves the current state of the game to JSON
+        /// </summary>
+        /// <param name="fileName"></param>
+        public void Save(string fileName)
+        {
+            Debug.Log("save initiated");
```
DavidPgl: Remove

Assets/Scripts/Managers/SaveManager.cs
```diff
@@ -0,0 +1,173 @@
...
+                save.campsites = new List<SavedCampsite>();
+                campsites.ForEach(cs => save.campsites.Add(new SavedCampsite(cs, cs.IsUnlocked)));
+
+                save.harvestables = sh;
+
+                Write(save, fileName);
+            }
+            catch (Exception e)
+            {
+                Debug.Log(e);
```
DavidPgl, line 74: Why is this a try catch?

Assets/Scripts/Managers/SaveManager.cs
```diff
@@ -0,0 +1,173 @@
...
+            }
+        }
+
+        /// <summary>
+        /// loads the savegame from the file & sets the states of gameobjects sequentially afterwards...
+        /// </summary>
+        /// <param name="fileName"></param>
+        public void Load(string fileName)
+        {
+            Debug.Log("loading initiated");
```
DavidPgl: Remove

Assets/Scripts/Managers/SaveManager.cs
```diff
@@ -0,0 +1,173 @@
...
+        //JSON - Utility functions - interaction with File system...
+
+        private void Write(Save save, string fileName)
+        {
+            string path = Application.persistentDataPath;
+            string json = JsonUtility.ToJson(save);
+            if (fileName == "") System.IO.File.WriteAllText(@"" + path + "/save.json", json);
+            else System.IO.File.WriteAllText(@"" + path + "/" + fileName + ".json", json);
+
+            Debug.Log("save written to file");
```
DavidPgl: Remove

Assets/Scripts/Utils/Saves/SavedHarvestable.cs
```diff
@@ -0,0 +1,24 @@
+﻿using Interactables;
+
+namespace Utils.Saves
+{
+    [System.Serializable]
+    public class SavedHarvestable
+    {
+        public Harvestable harvestable;
+        public bool isRespawning;
+        public double respawnTimePassed;
```
DavidPgl, line 10: All those fields could be made private with [SerializedField] for encapsulation-purposes, but if you don't change it I would be sad but I'll still approve it:)

DavidPgl (changes requested):

Assets/Scripts/Utils/Saves/SavedHarvestable.cs
inkompotato: They have to be public so that they correctly serialize to Json...

Assets/Scripts/Managers/SaveManager.cs
inkompotato: Writing to file might fail for unknown reasons (e.g. no write permissions)

Assets/Scripts/Interactables/Harvestable.cs
inkompotato: sure

Assets/Scripts/Buildings/BuildingBlueprint.cs
```diff
@@ -21,7 +21,11 @@ public class BuildingBlueprint : MonoBehaviour
         [Tooltip("How strong the blueprint should be faded out")] [SerializeField] [Range(0f, 1f)]
         private float fadeStrength;
 
-        public Building Building => building;
+        public Building Building
```
Kaes3kuch3n, line 24: Can be simplified to:
```suggestion
        public Building Building { get; set; }
```

Assets/Scripts/Entity/Player/PlayerState.cs
```diff
@@ -62,6 +53,35 @@ private void OnDisable()
             RemoteStatusHandler.OnPlayerSaturationRemoteUpdate -= ChangePlayerSaturation;
             RemoteStatusHandler.OnPlayerSanityRemoteUpdate -= ChangePlayerSanity;
         }
+        
+        public int GetHealth() => health;
```
Kaes3kuch3n, line 57: Getters and setters are fine, but since we are using C#, I think we could also use properties instead:
```c#
public int Health {
    get => health; 
    set {
        health = value;
        OnPlayerHealthUpdate?.Invoke(value);
    }
}
```

Assets/Scripts/Managers/SaveManager.cs
Kaes3kuch3n: But does anything else than the Write() statement need to be part of the try-block? 🤔 I think it would be better to pull the other stuff out of the try to make it clearer what can actually go wrong. You also might catch exceptions you actually don't wanna catch like this :)

Assets/Scripts/Managers/SaveManager.cs
```diff
@@ -0,0 +1,166 @@
...
+                    cs.SetState();
+                });
+
+                //set harvestables
+                save.harvestables.ForEach(h => { h.SetState(); });
+
+                // set inventory
+                inventoryController.SetItems(save.items);
+            }
+            catch (Exception e)
```
Kaes3kuch3n, line 137: See Save comment about try-catch :)

Assets/Scripts/Managers/SaveManager.cs
```diff
@@ -0,0 +1,166 @@
...
+            {
+                Debug.Log(e);
+            }
+        }
+
+
+        //JSON - Utility functions - interaction with File system...
+
+        private void Write(Save save, string fileName)
```
Kaes3kuch3n: I'd suggest adding an empty string as default for `fileName` for saving and loading :)
```suggestion
        private void Write(Save save, string fileName = "")
```

Assets/Scripts/Utils/Saves/SavedHarvestable.cs
DavidPgl: As far as i know, [SerializedField] should do that

inkompotato: > Generate a JSON representation of the public fields of an object.

from the Unity Documentation for JsonUtility.ToJson

inkompotato: therefore, [SerializeField] does not help when saving to json, the fields have to be public

Assets/Scripts/Entity/Player/PlayerState.cs
```diff
@@ -38,15 +38,6 @@ public class PlayerState : Damageable
...
         public static event PlayerIsDead OnPlayerDeath;
 
-        private void Awake()
-        {
-            // send event on initial values
-            OnPlayerHealthUpdate?.Invoke(health);
-            OnPlayerSaturationUpdate?.Invoke(saturation);
-            OnPlayerHydrationUpdate?.Invoke(hydration);
-            OnPlayerSanityUpdate?.Invoke(sanity);
-        }
```
SamuelllG, line 48: Please don't delete that, we need the events to be invoked on startup.

Assets/Scripts/Inventory/Equipable.cs
```diff
@@ -31,7 +31,7 @@ public class Equipable : Item
         /// <returns>Whether it has the given ability or not</returns>
         public bool HasAbility(EquipableAbility ability)
         {
-            if (_abilities == null)
+            if (_abilities.Count == 0)
```
SamuelllG, line 34: I think there's a difference between == null and .Count == 0, is that right?

Assets/Scripts/Managers/SaveManager.cs
```diff
@@ -0,0 +1,166 @@
...
+        public void Save(string fileName)
+        {
+            try
+            {
+                // grab required components to extract data from
+                GameObject player = PlayerManager.Instance.GetPlayer();
+                InventoryController inventoryController = InventoryManager.Instance.GetInvController();
+
+                // this is the player's current position in the map
+                Vector3 playerPosition = player.transform.localPosition;
```
SamuelllG, line 29: Why don't you use transform.position? Is there a difference in this case?

Assets/Scripts/Utils/Saves/Save.cs
```diff
@@ -0,0 +1,62 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Xml.Schema;
+using Buildings;
+using Inventory;
+using Managers;
+using UnityEngine;
+
+namespace Utils.Saves
```
SamuelllG, line 9: I think this isn't really a utility because it's only used for savegames, so maybe you could put these classes in their own folder.

Assets/Scripts/Utils/Saves/SavedHarvestable.cs
SamuelllG: Shouldn't public fields be capitalized, if they're not used in the inspector? Just to keep the consistency...

SamuelllG (dismissed):

Kaes3kuch3n (changes requested):

Assets/Scripts/Utils/Saves/SavedHarvestable.cs
Kaes3kuch3n: Since they could technically be used in the inspector, I think we should keep it like it ist currently 🤔 But it depends 😄

Assets/Scripts/Managers/SaveManager.cs
Kaes3kuch3n: I also think it should be `position` instead of `localPosition`

Assets/Scripts/Entity/Player/PlayerState.cs
inkompotato: nope, we do not - they only create errors as the events are invoked before everything has loaded correctly. On startup the default values (100 for all) are consistent with the UI. Changes will invoke the events again

Assets/Scripts/Inventory/Equipable.cs
inkompotato: there is but in this case the abilities array is never null, instead it might be an empty array

Assets/Scripts/Utils/Saves/SavedHarvestable.cs
inkompotato: as long as rider does not say anything about capitalization, it will not be capitalized.

Assets/Scripts/Utils/Saves/Save.cs
inkompotato: those are utility classes used for saving games. If you disagree, define Utility

Assets/Scripts/Buildings/BuildingBlueprint.cs
inkompotato: kk

Assets/Scripts/Managers/SaveManager.cs
inkompotato: kk

Assets/Scripts/Entity/Player/PlayerState.cs
inkompotato: since they are fine, no changes...

Assets/Scripts/Managers/SaveManager.cs
inkompotato: there is a lot that can go wrong, especially when you accidentally use an incompatible save. without the try-catch, the game might crash in such a case... therefore I would leave it in, just to be sure...

Assets/Scripts/Managers/SaveManager.cs
inkompotato: loading an incompatible save will result in a horrible amount of stuff that might go wrong, therefore try catch

Assets/Scripts/Buildings/BuildingBlueprint.cs
inkompotato: nope, this suggestion does not work and causes exceptions

Assets/Scripts/Managers/SaveManager.cs
inkompotato: the position that changes when the player moves is the localPosition. See the debug view in the Unity editor.

Assets/Scripts/Inventory/Equipable.cs
inkompotato: this is a fix suggested by david as it would have been broken before

Assets/Scripts/Buildings/BuildingBlueprint.cs
Kaes3kuch3n: Okay, nvm, my mistake 😅 Didn't recognise the backing field

Kaes3kuch3n (approved):

Assets/Scripts/Entity/Player/PlayerState.cs
SamuelllG: For the UI, yes but I put it there because other components needed it. What error do you mean? I didn't experience any

Assets/Scripts/Utils/Saves/Save.cs
SamuelllG: You put it to utility if the class is generally reusable (for other purposes) which savegames are not.

Assets/Scripts/Entity/Player/PlayerState.cs
inkompotato: Which other components?

inkompotato: as sanity should be unaffected with all player states at 100, it does not need to get a notification on startup, where all values are at 100

Assets/Scripts/Utils/Saves/Save.cs
inkompotato: they are reusable for anything that involves persistent data... resolved

DavidPgl (approved):
At last...

PR #165, opened 2020-01-24
Death scene update

tobihoffmann:
In addition to visual update of the death scene, there is a quickfix in PlayerState to prevent infinite invoking of OnPlayerDeath.

Also i changed the audiomanager while i was at it. It doesn't get destroyed anymore, this is essential since every scene uses it.


SamuelllG:
fixed attacking when target dead, removed quickfix

SamuelllG:
death screen won't work if game started via gamescene which doesn't really matter

please reopen @tobihoffmann

PR #169, opened 2020-01-25
Death scene update

tobihoffmann:
- deathscene new fonts/new model
- fixed bug in game scene where overlay fades in twice
- npc's now lose aggro when character dies
- audiomanager is now dontdestroyonload


Assets/Scripts/AbstractClasses/Damageable.cs
```diff
@@ -14,15 +14,23 @@ public abstract class Damageable : MonoBehaviour
...
             animator.SetTrigger(HitTrigger);
         }
 
+        /// <summary>
+        /// Does damage to the player. Implementation should call base.Hit(damage, attacker)
+        /// </summary>
+        /// <param name="damage">The damage dealth to the player</param>
+        /// <param name="health">The health of the entity after damage is dealt</param>
+        /// <param name="attacker">The EnemyController of the attacker if it's an NPC.</param>
+        public abstract void Hit(int damage, out int health, EnemyController attacker = null);
```
Kaes3kuch3n, line 32: I don't really like the fact that we have two Hit method definitions 😬 Wouldn't it be possible to just alter the existing implementation for your needs?

SamuelllG: I think this is the only possibility, if you don't want to create new references and coupling.

SamuelllG: AttackLogic and Hit is not clean anyway, since AttackLogic is used for both enemy and player. When we merge the saturation, hydration stuff that I'm working on, it won't get better unfortunately :(  But I don't have time to refactor it and rewrite the whole AttackLogic.

Kaes3kuch3n: Okay, it's fine then, I guess 🤷‍♂

Kaes3kuch3n (approved):

inkompotato (approved):

PR #171, opened 2020-01-25
Harvestable water

Kaes3kuch3n:


Kaes3kuch3n:
@potatoTVnet Could you check if this works with your save game logic? Shouldn't break anything, but just to make sure :)

inkompotato:
it is not covered by save logic, as there is a new script `Particle Harvestable`. Objects with that Type will not be queried by the SaveManager...

inkompotato (approved):

thuemmels (approved):

PR #176, opened 2020-01-25
Mo creatures

Kaes3kuch3n:


DavidPgl (dismissed):

DavidPgl (approved):

Assets/Scripts/Entity/Enemy/EnemyState.cs
```diff
@@ -45,11 +38,12 @@ private void Update()
             if (!targetFinder.HasTarget() && health < _maxHealth)
                 Regenerate();
         }
-        
+
         public override void Die()
         {
             base.Die();
-            _attackLogic.enabled = false;
+            if (_attackLogic)
```
SamuelllG, line 45: Either every enemy should have an AttackLogic or we have to make some changes (in EnemyController for example, nothing big, but something). The logic here implies that you could have enemies without AttackLogic.

SamuelllG (approved):

PR #178, opened 2020-01-26
Volume in settings

tobihoffmann:
- volume slider in settings does now work for ambience and music
- replaced stråy theme sketch with stråy theme final


thuemmels (changes requested):
Game throws exception when clicking on settings in the pause menu. 
(Volume slider seems a bit weird. In my opinion it is too quiet when set to 50%, compared to 100%)

tobihoffmann:
> Game throws exception when clicking on settings in the pause menu.

should be fine now.



> (Volume slider seems a bit weird. In my opinion it is too quiet when set to 50%, compared to 100%)

i know. this because ambience and music won't be played at 100% volume into the master channel. to fix it, we'd probably need different channels for music/ambience/effects and set the volumes in those specific channels. But this is a different issue.

thuemmels (approved):

DavidPgl (approved):

PR #183, opened 2020-01-26
decrease health when need == 0

SamuelllG:
you could improve the code by storing the ticks in an array, but I like to see which is which in the debug view 😅 therefore ...

### DIFF
--- a/Assets/Scripts/Entity/Player/NeedsController.cs
+++ b/Assets/Scripts/Entity/Player/NeedsController.cs
@@ -106,7 +106,7 @@ namespace Entity.Player
                     if (_playerState.Hydration <= 0)
                         _playerState.ChangePlayerHealth((int) (hydrationChange * healthDecreaseScale));
                     else
-                        _playerState.ChangePlayerSaturation(hydrationChange);
+                        _playerState.ChangePlayerHydration(hydrationChange);
                 }
             }
         }
--- a/Assets/Scripts/Entity/Player/NeedsController.cs
+++ b/Assets/Scripts/Entity/Player/NeedsController.cs
@@ -25,6 +25,10 @@ namespace Entity.Player
         [SerializeField]
         private float hydrationTick;
 
+        [Tooltip("A factor that scales the need tick and applies it on the player's health if the value of the need is at 0.")]
+        [SerializeField]
+        private float healthDecreaseScale;
+
         [Tooltip("A number that scales the saturation tick if moving.")]
         [SerializeField]
         private float moveMalusSaturation;
@@ -69,7 +73,7 @@ namespace Entity.Player
 
         private void Update()
         {
-            Tick();
+            TickAll();
             if (_movable.IsMoving())
                 _moveTimer += Time.deltaTime;
             else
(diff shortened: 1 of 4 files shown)


thuemmels (dismissed):

Assets/Scripts/Entity/Player/NeedsController.cs
```diff
@@ -69,30 +73,40 @@ private void Awake()
...
                 }
 
                 if (_nextHydrationTick >= 1)
                 {
-                    _playerState.ChangePlayerSaturation(GetNegativeChange(ref _nextHydrationTick));
+                    int hydrationChange = GetNegativeChange(ref _nextHydrationTick);
+                    if (_playerState.Hydration <= 0)
+                        _playerState.ChangePlayerHealth((int) (hydrationChange * healthDecreaseScale));
+                    else
+                        _playerState.ChangePlayerSaturation(hydrationChange);
```
Kaes3kuch3n: Not this time 😛

Kaes3kuch3n (changes requested):

Assets/Scripts/Entity/Player/NeedsController.cs
SamuelllG: I can't believe it 🤦‍♂

Kaes3kuch3n (approved):

Assets/Scripts/Managers/SaveManager.cs
```diff
@@ -96,10 +96,10 @@ public void Load(string fileName)
 
                 // set player state
                 PlayerState state = player.GetComponent<PlayerState>();
-                state.SetHealth(save.playerHealth);
-                state.SetHydration(save.playerHydration);
-                state.SetSanity(save.playerSanity);
-                state.SetSaturation(save.playerSaturation);
+                state.Health = save.playerHealth;
```
inkompotato, line 99: Have you tested this? Are Savegames still working as intended?

SamuelllG: I didn't test savegames..

inkompotato: As you made changes in the SaveManager, you should test it before we commit the changes...

Kaes3kuch3n: I think we should add some unit tests for this 🤔

SamuelllG: @Kaes3kuch3n haha :joy:
@potatoTVnet if you look at what I changed, you can see that the only thing I did, is making properties out of the getters and setters.. testing is always good, but in this case it wasn't worth the effort for me

SamuelllG: I didn't use savegames for now, I don't even have the newest version :( it would be nice if you could test if you want it to be tested

Kaes3kuch3n: Tested, still works fine @potatoTVnet

inkompotato (approved):

PR #195, opened 2020-01-27
Gitignore navmesh fix

SamuelllG:


inkompotato (approved):
while this PR should not have been necessary, it does fix the error and the Navmesh is fine again. The issue could have been prevented by actually checking out a PR that touches GameScene / NavMesh or similar files

cuongtm95 (approved):

SamuelllG:
this error was know although I didn't know it will break the whole build ^^

PR #196, opened 2020-01-27
Fix navmesh

SamuelllG:
this fix has to be two PRs, please approve

### DIFF
--- a/Assets/Scenes/GameScene/NavMesh.asset.meta
+++ b/Assets/Scenes/GameScene/NavMesh.asset.meta
@@ -1,8 +1,0 @@
-fileFormatVersion: 2
-guid: ef86375c5305c481689f838df46ed150
-NativeFormatImporter:
-  externalObjects: {}
-  mainObjectFileID: 23800000
-  userData: 
-  assetBundleName: 
-  assetBundleVariant: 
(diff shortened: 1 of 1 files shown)


DavidPgl (approved):

inkompotato (changes requested):
broken

inkompotato:
> broken

image

SamuelllG:
please approve jan ^^ we have to delete the navmesh, push it, and make a new one haha

SamuelllG:
Since changing the navmesh file causes it to break because the binary file isn't merged correctly, we have to delete it, and create it again. You can't do this in one PR because it then git recognizes the files as changed and we have the same problem.
(When the two PRs are on master and you checkout the branch, I guess it will still be broken, so you have to delete your local navmesh asset file before. But I'm not sure with this one.)

inkompotato:
you did update the navmesh in your previous PR #195 which worked

SamuelllG:
Yes! I updated it to not use the height mesh

SamuelllG:
Since we want the height mesh, that was a temporary solution

inkompotato:
ok, lets try it, revert if it fails

inkompotato (approved):

PR #197, opened 2020-01-27
create navmesh

SamuelllG:
Since changing the navmesh file causes it to break because the binary file isn't merged correctly, we have to delete it, and create it again. You can't do this in one PR because it then git recognizes the files as changed and we have the same problem.
(When the two PRs are on master and you checkout the branch, I guess it will still be broken, so you will probably have to delete your local navmesh asset file before pulling.)


Kaes3kuch3n (approved):

inkompotato (changes requested):
not working

SamuelllG:
Yep, you're right, then we have a bigger problem with git

inkompotato:
plz revert previous merges

SamuelllG:
Don't panic

PR #201, opened 2020-01-27
Load savegame from main menu

tobihoffmann:
Im not proud of it, but if anyone has a better solution, without completely ripping apart the SaveManager be my guest.


inkompotato:
There are issues that I do not understand: loading savegames works fine when starting the gamescene directly. or starting the game for the first time via menu (on master). The instanceID's used for saves change when returning to the main menu & going back into the game... This is not an issue with this commit but a result of unity being unity & our decision to use instanceID's for saving

inkompotato:
I could identify the following problem: 

Instances in the Gamescene do not get destroyed when returning to the main menu. Clicking new game or continue in the main menu results in a new game scene with new instances being loaded....

DavidPgl (dismissed):

Kaes3kuch3n (approved):

SamuelllG (approved):

PR #207, opened 2020-01-27
[Fixed multiple drops] [Fixed player being able to keep defiling corp…

DavidPgl:
…ses]

### DIFF
--- a/Assets/Scripts/Entity/Enemy/EnemyState.cs
+++ b/Assets/Scripts/Entity/Enemy/EnemyState.cs
@@ -41,7 +41,9 @@ namespace Entity.Enemy
 
         public override void Die()
         {
-            base.Die();
+            if (IsDead)
+                return;
+
             if (_attackLogic)
                 _attackLogic.enabled = false;
             Destroy(gameObject, 5f);
@@ -51,6 +53,8 @@ namespace Entity.Enemy
                     for (int i = 0; i < resourceData.amount; i++)
                         InventoryManager.Instance.AddItem(resourceData.item);
             });
+
+            base.Die();
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


SamuelllG (approved):
Didn't we implement something, so that mobs stop attacking when the player is dead? These two work together just fine?

DavidPgl:
> Didn't we implement something, so that mobs stop attacking when the player is dead? These two work together just fine?

Do you know where?

Assets/Scripts/Entity/Enemy/EnemyState.cs
```diff
@@ -41,7 +41,9 @@ private void Update()
 
         public override void Die()
         {
-            base.Die();
+            if (IsDead)
```
Kaes3kuch3n, line 44: I'd prefer it if you'd move this to the base method and keep `base.Die()` at the top here, since the check should be executed for all `Damageables` :)

Kaes3kuch3n (changes requested):

Kaes3kuch3n:
I've checked the stop attacking on entity death thingy, and it shouldn't be a problem. They work together perfectly fine

Assets/Scripts/Entity/Enemy/EnemyState.cs
DavidPgl: Can do, but then I have to find a nice way to tell the extended classes to skip as well

DavidPgl: Can do, but then I have to tell the extended classes to skip as well and I don't know how to do that nicely

Kaes3kuch3n: Okay, nvm, didn't think that one through 😅

Kaes3kuch3n (approved):

PR #227, opened 2020-01-29
Ui sound update

tobihoffmann:
Speechbubble has a sound now
Using consumables triggers a sound, depending on what they are


Kaes3kuch3n (approved):
uiiii

DavidPgl (approved):

PR #230, opened 2020-01-29
Adapts cam speed.

thuemmels:
Tbh, finding a balance between the right sensitivity for PC and laptop is tough, but this one is already much better than what we have right now. On my laptop it's fine now, on my PC it's maybe a bit faster than I would like but eh.

### DIFF
--- a/Assets/Prefabs/Entities/Player.prefab
+++ b/Assets/Prefabs/Entities/Player.prefab
@@ -242,7 +242,6 @@ MonoBehaviour:
       Boundary: 60
       Severity: 0.7
       Rate: -22.4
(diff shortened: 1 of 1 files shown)


SamuelllG:
Haha 0.05 is exactly what I found the best as well 👍 and who plays it on laptop anyway?? :smile:

SamuelllG (approved):

Kaes3kuch3n:
How about adding a slider to the settings?

SamuelllG:
Does that really make that much of a difference? 😅

Kaes3kuch3n (approved):

